Issue #12, opened 2020-09-15
API v8 support


vladfrangu:
If anyone wants to help out with this, be our guest! However this issue is more less for a TODO

- copy all files from v6 into a v8 folder
- update all payloads accordingly

zaida04:
I’d be willing to take a crack at this!

advaith1:
Here are the v8 changes:
> 1. permissions is now back to the original field, as a string.
> 2. game fields are now removed, replaced with the first element in activities. For widget endpoint, game field on users is renamed to activity, since there's still a singular field.
> 3. Channel Permission Overwrites have their type field serialized as a number (e.g. 0) instead of a string (e.g. role)
> 4. embed_enabled and embed_channel_id fields are removed, replaced with corresponding widget_ fields.
> 5. Form error responses now have a format that is better aligned with "regular errors".
> 6. Rate limit headers are sending seconds instead of milliseconds now. As such, the X-RateLimit-Precision header is no longer respected.
> 7. Bots no longer get CHANNEL_CREATE for DMs on v8.
> 8. Many routes have been deprecated.
> 9. Bans delete-message-days is no longer available. Use delete_message_days instead.
> 10. Removed roles, premium_since, and nick from Presence Update gateway event.

I'd also be willing to help if necessary

PR #13, opened 2020-09-15
chore: create v8 folder for future changes

monbrey:
As referenced in #12 

> copy all files from v6 into a v8 folder

This does that


monbrey:
#14 is much better, can collaborate on that if any further work needed

PR #14, opened 2020-09-15
feat: v8 support

### LINKED ISSUE #12: API v8 support

zaida04:
Accomplishes:
> - copy all files from v6 into a v8 folder
> - update all payloads accordingly

accompanied by the list of v8 changes (applicable ones at least):
> - permissions is now back to the original field, as a string.
> - game fields are now removed, replaced with the first element in activities. For widget endpoint, game field on users is renamed to activity, since there's still a singular field.
> - Channel Permission Overwrites have their type field serialized as a number (e.g. 0) instead of a string (e.g. role)
embed_enabled and embed_channel_id fields are removed, replaced with corresponding widget_ fields.
> - Form error responses now have a format that is better aligned with "regular errors".
> - Rate limit headers are sending seconds instead of milliseconds now. As such, the X-RateLimit-Precision header is no longer respected.
> - Bots no longer get CHANNEL_CREATE for DMs on v8.
> - Many routes have been deprecated.
> - Bans delete-message-days is no longer available. Use delete_message_days instead.
> - Removed roles, premium_since, and nick from Presence Update gateway event.


monbrey:
This doesn't close #12 yet - are there more commits on the way?

I'm not sure #12 should be closed at all and it looks like it's supposed to be an open TODO list anyway.

zaida04:
Oh, I meant when it’s NOT (WIP) it shall close that issue

zaida04:
More commits on the way, welcome to contr. on me fork

vladfrangu:
Please be aware of the changes in #15! They shouldn't conflict with the *current* state of this PR, however they should be merged into, then adapted to v8 wherever possible (if `_new` is dropped in responses, adapt accordingly) 👍

zaida04:
Is that it? Should I undraft it?

advaith1:
I think that's it but we should still wait for the docs PR next week

vladfrangu:
Can you also ensure any "Type"s are accurate? (example: Audit Log overwrite type)

advaith1:
@vladfrangu OverwriteType was updated, I don't think there's anything else

v8/payloads/auditLog.ts
```diff
@@ -0,0 +1,467 @@
...
+	 * - CHANNEL_OVERWRITE_UPDATE
+	 * - CHANNEL_OVERWRITE_DELETE
+	 *
+	 * **Present only if the {@link APIAuditLogOptions#type entry type} is "role"**
+	 */
+	role_name?: string;
+}
+
+export enum AuditLogOptionsType {
+	Member = 'member',
```
vladfrangu, line 158: Are these the same in v8?

advaith1: Oh, I didn't there's another type specifically for audit logs

... it looks like v8 sends those as a number in a string (`"1"`), I suppose that's a bug

advaith1: Opened an issue: https://github.com/discord/discord-api-docs/issues/2088

I guess we should leave this until it gets a reply

not sure why that needs to be a different type though? we could just set it to OverwriteType (if it should be just a number)

vladfrangu: If they're the same type.. and only for overwrites, you can delete it from both v6 and v8 and use the OverwriteType enum instead
First tho we have to wait for Discord's reply on the issue

advaith1:
https://github.com/discord/discord-api-docs/pull/2097

advaith1:
ok, I think this is done

v6/payloads/guild.ts
```diff
@@ -180,13 +180,21 @@ export interface APIGuildPreview {
 }
 
 /**
- * https://discord.com/developers/docs/resources/guild#guild-widget-object
+ * @deprecated Renamed to APIGuildWidgetSettings
```
vladfrangu, line 183:
```suggestion
 * @deprecated Use `APIGuildWidgetSettings` instead
```

v6/payloads/guild.ts
```diff
@@ -180,13 +180,21 @@ export interface APIGuildPreview {
...
  */
 export interface APIGuildWidget {
 	enabled: boolean;
 	channel_id: string | null;
 }
```
vladfrangu, line 188:
```suggestion
export type APIGuildWidget = APIGuildWidgetSettings;
```

v6/rest/guild.ts
```diff
@@ -337,17 +337,32 @@ export type RESTDeleteAPIGuildIntegrationResult = never;
...
+export type RESTPatchAPIGuildWidgetJSONBody = Partial<APIGuildWidgetSettings>;
 
 /**
  * https://discord.com/developers/docs/resources/guild#modify-guild-widget
  */
-export type RESTPatchAPIGuildWidgetJSONBody = Partial<APIGuildWidget>;
+export type RESTPatchAPIGuildWidgetSettingsJSONBody = Partial<APIGuildWidgetSettings>;
+
+/**
+ * @deprecated Renamed to RESTPatchAPIGuildWidgetSettingsResult
```
vladfrangu, line 361:
```suggestion
 * @deprecated Use `RESTPatchAPIGuildWidgetSettingsResult` instead
```

v6/rest/index.ts
```diff
@@ -351,12 +351,19 @@ export const Routes = {
 		return `/guilds/${guildID}/integrations/${integrationID}/sync`;
 	},
 
+	/**
+	 * @deprecated Renamed to guildWidgetSettings
```
vladfrangu, line 355:
```suggestion
	 * @deprecated Use `guildWidgetSettings` instead
```

v8/gateway/index.ts
```diff
@@ -0,0 +1,710 @@
...
+ */
+export type GatewayReadyDispatch = DataPayload<
+	GatewayDispatchEvents.Ready,
+	{
+		v: number;
+		user: APIUser;
+		session_id: string;
+		private_channels: [];
+		guilds: APIUnavailableGuild[];
+		shard?: [number, number];
```
vladfrangu, line 267:
```suggestion
		shard?: [shardID: number, shardCount: number];
```

vladfrangu (changes requested):
Some small nits

Also, if possible, adding a 
```md
@deprecated The types for the v6 version of the API types will not receive further updates. Please update to v8
```
to each v6 exported interface, objects, or whatnot, to mark that people should move to.. well, v8

vladfrangu:
One last note, other than the requested changes, please also keep in mind of https://github.com/discordjs/discord-api-types/pull/22 and integrate accordingly 🙏 (which was just merged 👀)

v6/rest/index.ts
```diff
@@ -9,7 +9,13 @@ export * from './user';
 export * from './voice';
 export * from './webhook';
 
+/**
+ * @deprecated API v6 is deprecated and the types will not receive further updates, please update to v8.
+ */
 export const APIVersion = '6';
+/**
```
vladfrangu, line 16:
```suggestion
export const APIVersion = '6';

/**
```

vladfrangu (changes requested):

vladfrangu (approved):

QuantumlyTangled (approved):

PR #22, opened 2020-09-21
fix(RESTPostAPIGuildsJSONBody): use correct types

NotSugden:
on the API docs for [Create Guild (POST /guilds)](https://discord.com/developers/docs/resources/guild#create-guild) there are a couple of info boxes stating that roles and channels can have an integer ID as a placeholder, eg for creating channel overwrites, this PR corrects these types into numbers

This PR also corrects the type for `roles` in `RESTPostAPIGuildsJSONBody` as it was previously typed as `APIRole` which isn't technically correct; the API won't accept permissons_new here, 

`afk_channel_id` and `system_channel_id` are documented to be of type snowflake, but testing shows that the API accepts an integer, (there is an issue on the API about this https://github.com/discord/discord-api-docs/issues/2089)

also i think these types need better names perhaps? im really not good with names so if anyones got suggestions

### DIFF
--- a/v6/rest/guild.ts
+++ b/v6/rest/guild.ts
@@ -7,6 +7,7 @@ import type {
 	APIGuildPreview,
 	APIGuildWidget,
 	APIInvite,
+	APIOverwrite,
 	APIRole,
 	APIVoiceRegion,
 	GuildDefaultMessageNotifications,
@@ -17,33 +18,22 @@ import type {
 	IntegrationExpireBehavior,
 } from '../payloads';
 
-export type APIGuildCreateOverwrite = Pick<
-	APIOverwrite,
-	'type'
-> & {
+export type APIGuildCreateOverwrite = Pick<APIOverwrite, 'type'> & {
 	id: string | number;
 	allow: number | string;
 	deny: number | string;
-}
+};
 
 export type APIGuildCreatePartialChannel = Partial<
-	Pick<
-		APIChannel,
-		'type' | 'topic' | 'nsfw' | 'bitrate' | 'user_limit' | 'rate_limit_per_user'
-	>
+	Pick<APIChannel, 'type' | 'topic' | 'nsfw' | 'bitrate' | 'user_limit' | 'rate_limit_per_user'>
 > & {
 	name: string;
 	id?: number;
 	parent_id?: number;
 	permission_overwrites?: APIGuildCreateOverwrite[];
-}
+};
 
-export type APIGuildCreateRole = Partial<
-	Pick<
-		APIRole,
-		'name' | 'color' | 'hoist' | 'position' | 'mentionable'
-	>
-> & {
+export interface APIGuildCreateRole extends RESTPostAPIGuildRoleJSONBody {
 	id?: number;
 	permissions?: number;
 }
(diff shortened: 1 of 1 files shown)


TTtie (approved):
Seems good to me

v6/rest/guild.ts
```diff
@@ -35,11 +48,11 @@ export interface RESTPostAPIGuildsJSONBody {
 	verification_level?: GuildVerificationLevel;
 	default_message_notifications?: GuildDefaultMessageNotifications;
 	explicit_content_filter?: GuildExplicitContentFilter;
-	roles?: APIRole[];
+	roles?: APIGuildCreateRole[];
 	channels?: APIGuildCreatePartialChannel[];
-	afk_channel_id?: string;
+	afk_channel_id?: number;
```
vladfrangu, line 53: Can't these be a number OR a string? That's how I understood it at least

vladfrangu (changes requested):

v6/rest/guild.ts
NotSugden: well, where would you get the string ID from? since the channels won't exist, although i remember trying this i think and i just got an invalid snowflake error from the API

vladfrangu: I mean you can generate IDs yourself, hence why I asked! If you could try it again, to confirm, that'd be awesome, and we can get this merged before v8 support 🎉

NotSugden: hm, it does seem you can pass in a snowflake as the placeholder IDs too, will update the PR to reflect that later

NotSugden: done, c36c7c3126e50276b448e4c0e735aea4f79dda36

v6/rest/guild.ts
```diff
@@ -17,13 +18,25 @@ import type {
...
-		APIChannel,
-		'type' | 'permission_overwrites' | 'topic' | 'nsfw' | 'bitrate' | 'user_limit' | 'rate_limit_per_user' | 'parent_id'
-	>
-> &
-	Required<Pick<APIChannel, 'name'>>;
+	Pick<APIChannel, 'type' | 'topic' | 'nsfw' | 'bitrate' | 'user_limit' | 'rate_limit_per_user'>
+> & {
+	name: string;
+	id?: number | string;
+	parent_id?: number;
```
vladfrangu, line 32:
```suggestion
	parent_id?: number | string;
```

v6/rest/guild.ts
```diff
@@ -17,13 +18,25 @@ import type {
 	IntegrationExpireBehavior,
 } from '../payloads';
 
+export type APIGuildCreateOverwrite = Pick<APIOverwrite, 'type'> & {
+	id: string | number;
```
vladfrangu, line 22:
```suggestion
	id: number | string;
```

v6/rest/guild.ts
vladfrangu, line 53:
```suggestion
	afk_channel_id?: number | string;
```

v6/rest/guild.ts
```diff
@@ -35,11 +48,11 @@ export interface RESTPostAPIGuildsJSONBody {
...
 	default_message_notifications?: GuildDefaultMessageNotifications;
 	explicit_content_filter?: GuildExplicitContentFilter;
-	roles?: APIRole[];
+	roles?: APIGuildCreateRole[];
 	channels?: APIGuildCreatePartialChannel[];
-	afk_channel_id?: string;
+	afk_channel_id?: number;
 	afk_timeout?: number;
-	system_channel_id?: string;
+	system_channel_id?: number;
```
vladfrangu, line 55:
```suggestion
	system_channel_id?: number | string;
```

vladfrangu (changes requested):
Last changes before merge!

vladfrangu (approved):

PR #24, opened 2020-10-04
feat(APIGuildWidgetMember): add activity and use proper status type

advaith1:


v8/payloads/guild.ts
```diff
@@ -204,7 +204,8 @@ export interface APIGuildWidgetMember {
 	username: string;
 	discriminator: string;
 	avatar: string | null;
-	status: string;
+	status: PresenceUpdateStatus;
+	activity: { name: string };
```
izexi, line 208: The `activity` prop isn't present when the member doesn't have a displayed activity.
```suggestion
	activity?: { name: string };
```

izexi (changes requested):

v8/payloads/guild.ts
advaith1: oh, I had fixed this but forgot to push lol

izexi (approved):

vladfrangu (approved):

PR #25, opened 2020-10-10
feat: server templates

advaith1:
https://github.com/discord/discord-api-docs/pull/1712
https://github.com/discord/discord-api-docs/pull/2144

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "discord-api-types",
-  "version": "0.5.0",
+  "version": "0.6.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/v8/payloads/template.ts
+++ b/v8/payloads/template.ts
@@ -1,0 +1,23 @@
+/**
+ * Types extracted from https://github.com/discord/discord-api-docs/blob/master/docs/resources/Template.md
+ */
+
+import type { APIUser } from './user'
+import type { APIGuild } from './guild'
+
+/**
+ * https://github.com/discord/discord-api-docs/blob/master/docs/resources/Template.md#template-structure
+ */
+export interface APITemplate {
+    code: string;
+    name: string;
+    description: string | null;
+    usage_count: number;
+    creator_id: string;
+    creator: APIUser;
+    created_at: string;
+    updated_at: string;
+    source_guild_id: string;
+    serialized_source_guild: Partial<APIGuild>;
+    is_dirty: boolean | null;
+}
--- a/v8/payloads/template.ts
+++ b/v8/payloads/template.ts
@@ -3,7 +3,7 @@
  */
 
 import type { APIUser } from './user';
-import type { APIGuild } from './guild';
+import type { RESTPostAPIGuildsJSONBody } from '../rest';
 
 /**
  * https://discord.com/developers/docs/resources/template#template-object
@@ -18,6 +18,12 @@ export interface APITemplate {
 	created_at: string;
 	updated_at: string;
 	source_guild_id: string;
-	serialized_source_guild: Partial<APIGuild>;
+	serialized_source_guild: APITemplateSerializedSourceGuild;
 	is_dirty: boolean | null;
+}
+
+export interface APITemplateSerializedSourceGuild extends Omit<RESTPostAPIGuildsJSONBody, 'icon'> {
+	description: string | null;
+	preferred_locale: string;
+	icon_hash: string | null;
 }
--- a/v8/payloads/template.ts
+++ b/v8/payloads/template.ts
@@ -1,12 +1,12 @@
 /**
- * Types extracted from https://github.com/discord/discord-api-docs/blob/master/docs/resources/Template.md
+ * Types extracted from https://discord.com/developers/docs/resources/template
  */
 
 import type { APIUser } from './user';
 import type { APIGuild } from './guild';
 
 /**
- * https://github.com/discord/discord-api-docs/blob/master/docs/resources/Template.md#template-structure
+ * https://discord.com/developers/docs/resources/template#template-object
  */
 export interface APITemplate {
 	code: string;
--- a/v8/payloads/template.ts
+++ b/v8/payloads/template.ts
@@ -2,22 +2,22 @@
  * Types extracted from https://github.com/discord/discord-api-docs/blob/master/docs/resources/Template.md
  */
 
-import type { APIUser } from './user'
-import type { APIGuild } from './guild'
+import type { APIUser } from './user';
+import type { APIGuild } from './guild';
 
 /**
  * https://github.com/discord/discord-api-docs/blob/master/docs/resources/Template.md#template-structure
  */
 export interface APITemplate {
-    code: string;
-    name: string;
-    description: string | null;
-    usage_count: number;
-    creator_id: string;
-    creator: APIUser;
-    created_at: string;
-    updated_at: string;
-    source_guild_id: string;
-    serialized_source_guild: Partial<APIGuild>;
-    is_dirty: boolean | null;
+	code: string;
+	name: string;
+	description: string | null;
+	usage_count: number;
+	creator_id: string;
+	creator: APIUser;
+	created_at: string;
+	updated_at: string;
+	source_guild_id: string;
+	serialized_source_guild: Partial<APIGuild>;
+	is_dirty: boolean | null;
 }
--- a/v8/rest/index.ts
+++ b/v8/rest/index.ts
@@ -5,7 +5,7 @@ export * from './gateway';
 export * from './guild';
 export * from './invite';
 export * from './oauth2';
-export * from './template'
+export * from './template';
 export * from './user';
 export * from './voice';
 export * from './webhook';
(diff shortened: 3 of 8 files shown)


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "discord-api-types",
-  "version": "0.5.0",
+  "version": "0.6.0",
```
vladfrangu, line 3: Ha, no.
```suggestion
  "version": "0.5.1",
```

v8/rest/index.ts
```diff
@@ -393,6 +394,34 @@ export const Routes = {
...
+	template(code: string) {
+		return `/guilds/templates/${code}`;
+	},
+
+	/**
+	 * Route for:
+	 * - GET `/guilds/{guild.id}/templates`
+	 * - POST `/guilds/{guild.id}/templates`
+	 */
+	guildTemplate(guildID: string) {
```
vladfrangu, line 411: This should probably be `guildTemplates`

v8/rest/index.ts
```diff
@@ -393,6 +394,34 @@ export const Routes = {
...
+		return `/guilds/${guildID}/templates`;
+	},
+
+	/**
+	 * Route for:
+	 * - PUT `/guilds/{guild.id}/templates/{template.code}`
+	 * - PATCH `/guilds/{guild.id}/templates/{template.code}`
+	 * - DELETE `/guilds/{guild.id}/templates/{template.code}`
+	 */
+	guildTemplateModify(guildID: string, code: string) {
```
vladfrangu, line 421: In contrast to above, `guildTemplate`

v8/rest/template.ts
```diff
@@ -0,0 +1,34 @@
+import type { APIGuild, APITemplate } from '../payloads';
+
+/**
+ * https://github.com/discord/discord-api-docs/blob/master/docs/resources/Template.md#get-template--get-guildstemplatestemplatecodedocs_resources_templatetemplate-object
+ */
```
vladfrangu, line 5: `REST<Method>APIGuildTemplate` please 🙏

vladfrangu: And what's the difference between this and `RESTGetAPIGuildTemplateResult`?

v8/rest/template.ts
```diff
@@ -0,0 +1,34 @@
...
+export interface RESTPostAPIGuildTemplateJSONBody {
+	name: string;
+	description?: string | null;
+}
+
+export type RESTPostAPIGuildTemplateResult = APITemplate;
+
+export type RESTPutAPIGuildTemplateResult = APITemplate;
+
+export type RESTPatchAPIGuildTemplateJSONBody = RESTPostAPIGuildTemplateJSONBody;
```
vladfrangu, line 31:
```suggestion
export type RESTPatchAPIGuildTemplateJSONBody = Partial<RESTPostAPIGuildTemplateJSONBody>;
```

vladfrangu (changes requested):

package-lock.json
advaith1: npm automatically changed this because it's 0.6.0 in package.json

vladfrangu: Oh pfft, it's the lock file
I'm blind

v8/rest/template.ts
advaith1: different endpoints, but the other one should be an array oops

v8/rest/index.ts
advaith1: i documented it as template because you can actually only have one template per guild, it was changed before release so the routes are kinda inconsistent with the behaviour

v8/rest/template.ts
advaith1: https://github.com/discord/discord-api-docs/pull/2144#discussion_r502833612

v8/rest/index.ts
vladfrangu: Still, naming usually follows its route counterpart 👁️

v8/rest/template.ts
vladfrangu: If the data returned is the same, it should.. probably be unified, and exported as an alias

advaith1: you're talking about `export type RESTGetAPITemplateResult = APITemplate;` right? (because this comment is attached to a comment line)
that route returns something else and I edited it, but this route is just the APITemplate object

v8/payloads/template.ts
```diff
@@ -0,0 +1,23 @@
...
+	code: string;
+	name: string;
+	description: string | null;
+	usage_count: number;
+	creator_id: string;
+	creator: APIUser;
+	created_at: string;
+	updated_at: string;
+	source_guild_id: string;
+	serialized_source_guild: Partial<APIGuild>;
```
vladfrangu, line 21: Apparently..this is the same as the `RESTPostAPIGuildCreateJSONBody`?

v8/rest/template.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * https://github.com/discord/discord-api-docs/blob/master/docs/resources/Template.md#get-template--get-guildstemplatestemplatecodedocs_resources_templatetemplate-object
+ */
+export type RESTGetAPITemplateResult = APITemplate;
+
+/**
+ * https://github.com/discord/discord-api-docs/blob/master/docs/resources/Template.md#create-guild-from-template--post-guildstemplatestemplatecodedocs_resources_templatetemplate-object
+ */
+export interface RESTPostAPITemplateJSONBody {
```
vladfrangu, line 11: So this is what... create guild with template? If so, this should..really be renamed to make it clearer 🙏  (something like `RESTPostAPIGuildCreateWithTemplate` or something.. it's an edge case where name doesn't match route

v8/rest/template.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export type RESTGetAPIGuildTemplatesResult = APITemplate[];
+
+export interface RESTPostAPIGuildTemplatesJSONBody {
+	name: string;
+	description?: string | null;
+}
+export type RESTPostAPIGuildTemplatesResult = APITemplate;
+
+export type RESTPutAPIGuildTemplateResult = APITemplate;
```
vladfrangu, line 28:
```suggestion
export type RESTPutAPIGuildTemplateJSONBody = RESTPostAPIGuildTemplatesJSONBody;
export type RESTPutAPIGuildTemplateResult = APITemplate;
```

vladfrangu (changes requested):

v8/rest/template.ts
advaith1: the PUT req doesn't take a body

v8/payloads/template.ts
advaith1: no? serialized_source_guild has a lot more than those 2 fields ([example](https://github.com/discord/discord-api-docs/blob/master/docs/resources/Template.md#example-template-object))

v8/rest/template.ts
vladfrangu: Oh it's `sync`...Riiight

v8/payloads/template.ts
vladfrangu: Which are also present in `RESTPostAPIGuildCreateJSONBody`, which has a LOT of fields (like roles, channels, etc...) and it was mentioned in some of the PRs you linked

vladfrangu: See https://github.com/discord/discord-api-docs/pull/2144#issuecomment-708490998

advaith1: oh sorry i misread, so use that but replace icon with icon_hash? looks like it also has a `description` field though

vladfrangu: Is..the JSON body for create guild outdated? Or are there values only for templates? 🤔

advaith1: `icon_hash` is an exception apparently

`description` isn't really useful in creating guilds as it's only for discoverable (or preview_enabled) guilds

vladfrangu: Useful or not, we should document it if it can be used 👍

advaith1: yeah so it should stay APIGuild, since that's the one with the description property and that's what it's documented as

vladfrangu: Yeah..I don't know what to say about that really.. Feedback needed for this one.

Discord claims `serialized_source_guild` is exactly the data that the `Create Guild` endpoint takes (in this case `RESTPostAPIGuildCreateJSONBody` which also accounts for IDs being numbers (!)). Per https://github.com/discord/discord-api-docs/pull/2144#issuecomment-708490998, seems like icon_hash isn't even used anymore (although it's still set... neat). While typing this as `Partial<APIGuild>` isn't necessarily *wrong*, it's also not *right*...

CC @kyranet @iCrawl what do you say?

kyranet: We can still make a `APITemplateSerializedSourceGuildResult`, and if the fields are more similar to `RESTPostAPIGuildCreateJSONBody` but the properties differ, we can omit the ones missing and add the ones that aren't.

vladfrangu: As it stands, `RESTPostAPIGuildCreateJSONBody` is most accurate.. a small `type APITemplateSerializedSourceGuild = RESTPostAPIGuildCreateJSONBody & { whateverIsMissing }` is plenty

vladfrangu: @advaith1 ^ since your PR(s?) were merged

vladfrangu (approved):

kyranet (approved):

PR #30, opened 2020-10-18
fix(GatewayPresence): correct type for sent activity objects

Vaporox:
[API Docs](https://discord.com/developers/docs/topics/gateway#activity-object-activity-structure): Bots are only able to send `name`, `type`, and optionally `url`.

There is no separation between activity objects received from Discord and ones sent from the client. That causes problems like this, even though `created_at` should not be sent / will be ignored anyway.

image

I'm not fully familiar with the file structure yet, so if I placed it where it doesn't belong just let me know.


vladfrangu (approved):
I'll get it right eventually:tm:

*curse you Discord with confusing types*

PR #33, opened 2020-10-18
chore: fix mistypes (literally) for inline replies

vladfrangu:


v8/payloads/channel.ts
```diff
@@ -102,7 +102,7 @@ export enum MessageType {
 	GUILD_DISCOVERY_REQUALIFIED,
 	GUILD_DISCOVERY_GRACE_PERIOD_INITIAL_WARNING,
 	GUILD_DISCOVERY_GRACE_PERIOD_FINAL_WARNING,
-	INLINE_REPLY,
+	INLINE_REPLY = 18,
```
advaith1, line 105: this is already implied

vladfrangu: Woops, meant to make it 19, good catch

kyranet (approved):

PR #34, opened 2020-10-28
feat(Message): reply updates

advaith1:
whether a reply should ping the author of the original message


vladfrangu (approved):
Waiting on Discord to either get a PR or mention this somewhere, otherwise LGTM

advaith1:
it's in the replies pr now

but now channel_id in allowed 
_mentions is optional so I can add that to this pr

vladfrangu (changes requested):
Before we can merge this, https://github.com/discordjs/discord-api-types/pull/34/files#diff-df009aedf43eb83e065b2a738b64184c0e6d5fd592a9d9216be0354c761d0a2fR79 needs an update because the messageID would be required if you are sending a message_reference

advaith1:
so should we make a different interface for sending? in docs it's the same object

vladfrangu:
No.. `Required<APIMessageReference> & Omit<APIMessageReference, 'message_id'> ` or similar

vladfrangu (approved):

vladfrangu (approved):

PR #58, opened 2020-12-25
fix(GatewayActivityEmoji): fix types for `name` and `id`

cherryblossom000:
According to the [docs](https://discord.com/developers/docs/topics/gateway#activity-object-activity-emoji) `name` is required and `id` is optional. Both are also non-nullable.

Previously, the typings incorrectly had `name` as optional and `id` as required, with both being nullable.

### DIFF
--- a/v8/payloads/gateway.ts
+++ b/v8/payloads/gateway.ts
@@ -109,7 +109,10 @@ export interface GatewayActivityTimestamps {
 /**
  * https://discord.com/developers/docs/topics/gateway#activity-object-activity-emoji
  */
-export type GatewayActivityEmoji = Partial<Pick<APIEmoji, 'name' | 'animated'>> & Pick<APIEmoji, 'id'>;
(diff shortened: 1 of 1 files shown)


v8/payloads/gateway.ts
```diff
@@ -109,7 +109,10 @@ export interface GatewayActivityTimestamps {
 /**
  * https://discord.com/developers/docs/topics/gateway#activity-object-activity-emoji
  */
-export type GatewayActivityEmoji = Partial<Pick<APIEmoji, 'name' | 'animated'>> & Pick<APIEmoji, 'id'>;
+export interface GatewayActivityEmoji extends Pick<APIEmoji, 'animated'> {
```
vladfrangu, line 112: Can you keep it like it was before, only make the od part of the Partial?

vladfrangu (changes requested):

vladfrangu:
This fix has been added to #62 (unintentionally and indirectly)... Will credit you for them tho! 😅

PR #62, opened 2020-12-27
chore(chores): All the chores! And release 0.12.0

vladfrangu:
Basically take PRs, or issues, and do them all at once!

Major version time!

---

#### Credits

- https://github.com/discordjs/discord-api-types/pull/58 by @cherryblossom000 for GatewayActivityEmoji fix


v8/payloads/auditLog.ts
```diff
@@ -19,22 +19,72 @@ import type { APIWebhook } from './webhook';
  * https://discord.com/developers/docs/resources/audit-log#audit-log-object-audit-log-structure
  */
 export interface APIAuditLog {
+	/**
+	 * Webhooks found in the audit log
+	 *
+	 * See https://discord.com/developers/docs/resources/webhook#webhook-object
```
kyranet, line 25: This should use `@see`, now that TypeScript 4.1 [added support for it](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#jsdoc-see-tag), same for everything else.

kyranet (dismissed):

v8/payloads/auditLog.ts
vladfrangu: Intentionally chose against it so that the link is actually clickable in VSC.

v8/gateway/index.ts
```diff
@@ -26,78 +26,240 @@ export const GatewayVersion = '8';
...
+	 * Acknowledge a successful session resume.
+	 */
 	Resumed,
-
+	/**
+	 * A client has connected to the voice channel
+	 */
 	ClientConnect = 12,
+	/**
+	 * A client has disconnected from the voice channel
```
NotSugden, line 207:
```suggestion
	 * A client has disconnected from the voice channel.
```

v8/gateway/index.ts
```diff
@@ -26,78 +26,240 @@ export const GatewayVersion = '8';
...
+	 * Time to wait between sending heartbeats in milliseconds.
+	 */
 	Hello,
+	/**
+	 * Acknowledge a successful session resume.
+	 */
 	Resumed,
-
+	/**
+	 * A client has connected to the voice channel
```
NotSugden, line 203:
```suggestion
	 * A client has connected to the voice channel.
```

v8/gateway/index.ts
```diff
@@ -229,6 +391,9 @@ export interface GatewayHello extends NonDispatchPayload {
  * https://discord.com/developers/docs/topics/gateway#hello
  */
 export interface GatewayHelloData {
+	/**
+	 * The interval (in milliseconds) the client should heartbeat with
```
NotSugden, line 395:
```suggestion
	 * The interval (in milliseconds) the client should heartbeat with.
```

v8/gateway/index.ts
```diff
@@ -278,12 +443,43 @@ export type GatewayReadyDispatch = DataPayload<GatewayDispatchEvents.Ready, Gate
  * https://discord.com/developers/docs/topics/gateway#ready
  */
 export interface GatewayReadyDispatchData {
+	/**
+	 * Gateway version
```
NotSugden, line 447:
```suggestion
	 * Gateway version.
```

v8/gateway/index.ts
```diff
@@ -278,12 +443,43 @@ export type GatewayReadyDispatch = DataPayload<GatewayDispatchEvents.Ready, Gate
...
+	 *
+	 * See https://discord.com/developers/docs/topics/gateway#gateways-gateway-versions
+	 */
 	v: number;
+	/**
+	 * Information about the user including email
```
NotSugden, line 453:
```suggestion
	 * Information about the user including email.
```

NotSugden (commented):
are the full stops meant to be on every jsdoc? or only specific ones? or.. ?
the first few comments include them, but the rest do not

package.json
```diff
@@ -1,94 +1,125 @@
...
+	},
+	"husky": {
+		"hooks": {
+			"pre-commit": "lint-staged",
+			"commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
+		}
+	},
+	"lint-staged": {
+		"*.js": "eslint --fix",
+		"*.ts": "eslint --fix --ext ts"
```
Favna, line 96:
```suggestion
		"*.{mjs,js,ts}": "eslint --fix --ext mjs,js,ts"
```

This can be 1 line. Also since ESLint is configured to lint mjs, should also include it here.

scripts/deno.mjs
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
+import { readdir, mkdir, readFile, writeFile, rm, stat, copyFile } from 'fs/promises';
+import { dirname, join } from 'path';
+import { fileURLToPath } from 'url';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+const baseDirectory = join(__dirname, '..');
+const denoPath = join(baseDirectory, '.deno');
```
Favna, line 10: Unless you insist on using `join`, Kyra would ree at this approach as opposed to 

```suggestion
const baseDirectory = new URL('../', import.meta.url);
const denoPath = new URL('.deno/', baseDirectory);
```

Note that if you still need the regular file path at a later stage you can always call:
```mjs
import { fileURLToPath } from 'url';
fileURLToPath(urlThatIsAFileOrDirectoryHere),
```
However, all `fs/promises` functions take a `PathLike` (from `@types/node/fs.d.ts`):
```ts
export type PathLike = string | Buffer | URL;
```

tsconfig.json
```diff
@@ -15,15 +15,6 @@
...
-		"bin",
-		"examples",
-		"scripts"
-	],
-	"include": [
-		"v*/**/*",
-		"default/**/*",
-		"common/**/*",
-	]
+	"exclude": ["node_modules"],
```
Favna, line 18:
```suggestion
```

Unless you expect any of the include patterns to have `node_modules` this does nothing.

From: https://www.typescriptlang.org/tsconfig#exclude

> Important: `exclude ` *only* changes which files are included as a result of the `include` setting. A file specified by exclude can still become part of your codebase due to an import statement in your code, a `types` inclusion, a `/// <reference` directive, or being specified in the `files` list.

SpaceEEC (approved):

Favna (approved):

PR #77, opened 2021-01-27
chore: miscellaneous tasks

vladfrangu:
~~With this PR, we are moving from `0.x` major releases to `1.x.x` major releases. No big changes, your code will work as is! Still debating if we should move to this or stay at `0.x` majors, but both would yield the same result~~ Decision was made - sticking to `0.x` 

Otherwise, this PR is a draft while I.. draft up the release and check how up to date we are with Discord's documentation


kyranet (approved):

v8/rest/index.ts
```diff
@@ -638,7 +638,7 @@ export const OAuth2Routes = {
 } as const;
 
 /**
- * Freeze route object
+ * Freeze Oauth2 route object
```
iCrawl, line 641: OAuth2

iCrawl (approved):

PR #89, opened 2021-02-13
fix(deno): replace `const enum` exports in deno with normal `enum`s

Apacheli:
AFAIA, Deno does not support `const enum` due to `--isolatedModules` being enabled https://github.com/denoland/deno/discussions/9082#discussioncomment-273198 (possible TypeScript bug)

I have added fixes although I want to be sure that this works. Any Deno reviewers would be extra appreciated.

### DIFF
--- a/deno/common/mod.ts
+++ b/deno/common/mod.ts
@@ -1,7 +1,7 @@
 /**
  * https://discord.com/developers/docs/topics/opcodes-and-status-codes#json-json-error-codes
  */
-export const enum RESTJSONErrorCodes {
+export enum RESTJSONErrorCodes {
 	GeneralError,
 
 	UnknownAccount = 10001,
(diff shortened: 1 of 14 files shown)


Apacheli:
Side-note: After making this PR, I realized v6 had this fixed. Maybe someone forgot to change them for v8?

vladfrangu:
1. v6 hasn't and isn't getting any updates anymore, hence why it didn't get the `const enum` treatment
2. Since deno likes being funky and doesn't like const enums, the proper fix would be replacing all const enums with normal ones in deno exports, at least till they support const ones (if they will that is). I can PR that later today or tomorrow! :D

kyranet (approved):

Apacheli:
> 
> 
>     1. v6 hasn't and isn't getting any updates anymore, hence why it didn't get the `const enum` treatment
> 
>     2. Since deno likes being funky and doesn't like const enums, the proper fix would be replacing all const enums with normal ones in deno exports, at least till they support const ones (if they will that is). I can PR that later today or tomorrow! :D

Sorry for the ***incredibly*** late reply but I think this PR is ready.

Apacheli:
Weird, creating a branch from the merge conflict might have ruined the original PR. I will need to fix that

vladfrangu:
I'll fix it up :D
I was gonna push what I meant anyways for the script

vladfrangu (approved):

PR #107, opened 2021-04-01
feat: stage channels!

vladfrangu:

Stage channels!

https://github.com/discord/discord-api-docs/pull/2751

### DIFF
--- a/deno/v8/payloads/channel.ts
+++ b/deno/v8/payloads/channel.ts
@@ -108,7 +108,7 @@ export enum ChannelType {
 	/**
 	 * A text channel within a guild
 	 */
-	GUILD_TEXT = 0,
+	GUILD_TEXT,
 	/**
 	 * A direct message between users
 	 */
@@ -139,6 +139,12 @@ export enum ChannelType {
 	 * See https://discord.com/developers/docs/game-and-server-management/special-channels
 	 */
 	GUILD_STORE,
+	/**
+	 * A voice channel for hosting events with an audience
+	 *
+	 * See https://support.discord.com/hc/en-us/articles/1500005513722
+	 */
+	GUILD_STAGE_VOICE = 13,
 }
 
 /**
--- a/v8/rest/guild.ts
+++ b/v8/rest/guild.ts
@@ -770,3 +770,29 @@ export interface RESTPatchAPIGuildMemberVerificationJSONBody {
 }
 
 export type RESTPatchAPIGuildMemberVerificationResult = APIGuildMembershipScreening;
+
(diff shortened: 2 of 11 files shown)


v8/rest/guild.ts
```diff
@@ -770,3 +770,29 @@ export interface RESTPatchAPIGuildMemberVerificationJSONBody {
...
 
 export type RESTPatchAPIGuildMemberVerificationResult = APIGuildMembershipScreening;
+
+export interface RESTPatchAPIGuildVoiceStateCurrentMemberJSONBody {
+	/**
+	 * The id of the channel the user is currently in
+	 */
+	channel_id: Snowflake;
+	/**
+	 * Toggles the user's suppress state
```
amishshah, line 780: I think this is more accurate than "toggles"
```suggestion
	 * Whether the user is suppressed from speaking
```

v8/rest/guild.ts
```diff
@@ -770,3 +770,29 @@ export interface RESTPatchAPIGuildMemberVerificationJSONBody {
...
+	 */
+	suppress?: boolean;
+	/**
+	 * Sets the user's request to speak
```
amishshah, line 784:
```suggestion
	 * The time at which the user requested to speak (ISO8601 timestamp)
```

amishshah (changes requested):
I had 2 changes that I thought improved clarity on the two new voice state properties.

deno/v8/payloads/channel.ts
```diff
@@ -139,6 +139,12 @@ export enum ChannelType {
 	 * See https://discord.com/developers/docs/game-and-server-management/special-channels
 	 */
 	GUILD_STORE,
```
amishshah, line 141: Also, I think it would be worth explicitly setting values for the other channel types to avoid some accidental mistakes. It would also appear more consistent with the STAGE_CHANNEL line.

amishshah (changes requested):

v8/rest/guild.ts
vladfrangu: I copied the description they have in the PR, and that's what's been done for forever now...even tho at times their description is..iffy

amishshah: Fair enough!

deno/v8/payloads/channel.ts
vladfrangu: Usually for enums, the only thing I do is set the value if it's a jump, and not explicitly incremental (see JSONErrorCodes)... It's simpler too :3

amishshah (approved):

kyranet (approved):

PR #113, opened 2021-04-12
fix(TypeScript): imports not working in TypeScript

vladfrangu:

While the node.js exports work perfectly fine, TypeScript doesn't parse those.
As such, TS users would get an error when importing the module.
Alongside that, this commit fixes eslint for mjs files,
which need @babel to process TLA

N/A

### DIFF
--- a/scripts/versions.mjs
+++ b/scripts/versions.mjs
@@ -9,7 +9,7 @@ const rootDir = join(__dirname, '..');
 const execAsync = promisify(exec);
 
 const fileToESMWrapperCall = (path, version) =>
-	execAsync(`npx gen-esm-wrapper ${rootDir}${path}/${version}.js ${rootDir}${path}/${version}.mjs`);
+	execAsync(`npx gen-esm-wrapper "${join(rootDir, path, `${version}.js`)}" "${join(rootDir, path, `${version}.mjs`)}"`);
 
 await Promise.allSettled(
 	[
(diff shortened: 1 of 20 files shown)


scripts/versions.mjs
```diff
@@ -29,7 +29,7 @@ await Promise.allSettled(
 			fileToESMWrapperCall('utils', version),
 
 			// Shortcuts
-			fileToESMWrapperCall('shortcuts', version),
+			fileToESMWrapperCall('owo/..', version),
```
kyranet, line 32: What's this

vladfrangu: Joke: OwO
Actual: The script expects a path to the file to convert, and you cannot do `./../` sadly (I tried). So, to alleviate, I gave it a non-existent path just so it can go to the root folder where the file is

kyranet: Then use [fileURLToPath](https://nodejs.org/dist/latest-v15.x/docs/api/url.html#url_url_fileurltopath_url), it's less hacky and should work if this is just expecting a path:

```javascript
const __filename = fileURLToPath(import.meta.url);
const __dirname = dirname(__filename);
const rootDir = join(__dirname, '..');
```

If that doesn't, you should refactor the path checking to check whether [it's absolute](https://nodejs.org/dist/latest-v15.x/docs/api/path.html#path_path_isabsolute_path) or not.

vladfrangu: Done in https://github.com/discordjs/discord-api-types/pull/113/commits/1fa258ed3f36d4d786fb4ebca0a55b89a7d5b788

scripts/versions.mjs
```diff
@@ -1,10 +1,15 @@
...
 import { promisify } from 'node:util';
 
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+const rootDir = join(__dirname, '..');
 const execAsync = promisify(exec);
 
 const fileToESMWrapperCall = (path, version) =>
-	execAsync(`npx gen-esm-wrapper ./${path}/${version}.js ./${path}/${version}.mjs`);
+	execAsync(`npx gen-esm-wrapper ${rootDir}${path}/${version}.js ${rootDir}${path}/${version}.mjs`);
```
kyranet, line 12: You could use `join` for building OS-normalized paths, and you should wrap them between quotes, or else this script will fail when the project is inside a folder with spaces.

kyranet (changes requested):

kyranet (approved):

PR #129, opened 2021-05-02
feat: add new interfaces for interaction-related structures

vaporox:

This adds new interfaces for the guild members and channels returned when dealing with interactions. Previously, the extra properties were just added to the base interface, but that way it's harder to import and use them. (This will also be useful for discordjs/discord.js#5448)

### DIFF
--- a/deno/payloads/v8/interactions.ts
+++ b/deno/payloads/v8/interactions.ts
@@ -100,21 +100,6 @@ export interface APIApplicationCommandOptionChoice {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


vladfrangu (approved):

vaporox:
@vladfrangu I refactored it a bit so that the interaction member and resolved members do not share the same interface, it would be good if you could have another look

vladfrangu:
I mean.. sure, that works too!

amishshah (approved):

PR #133, opened 2021-05-21
feat: api v9 and threads

ckohen:

This draft documents API / Gateway v9 and threads.
This also adds the `thread_id` property on execute webhook.
Obviously the data here is not finalized on the API side, but it seems like relatively few changes are coming so now the majority of the work is done.

BREAKING CHANGE: default exports are now v9

I *think* I hit everything, but I may have missed a few subtleties.

https://github.com/discord/discord-api-docs/pull/2855
https://github.com/discord/discord-api-docs/pull/2875


payloads/v9/channel.ts
```diff
@@ -0,0 +1,1047 @@
...
+	Role,
+	Member,
+}
+
+/**
+ * https://discord.com/developers/docs/resources/channel#thread-metadata-object-thread-metadata-structure
+ */
+export interface APIThreadMetadata {
+	/**
+	 * The archived state of the thread
```
vladfrangu, line 616: These should use the same strings that the docs use

See https://discord.com/developers/docs/resources/channel#thread-metadata-object

payloads/v9/channel.ts
```diff
@@ -0,0 +1,1047 @@
...
+	 * An ISO8601 timestamp for when `archived` was last changed
+	 */
+	archive_timestamp: string;
+	/**
+	 * The locked state of the thread
+	 */
+	locked?: boolean;
+}
+
+export type ThreadAutoArchiveDuration = 60 | 1440 | 4320 | 10080;
```
vladfrangu, line 637: let's make this an enum instead, easier for the end users

payloads/v9/channel.ts
```diff
@@ -0,0 +1,1047 @@
...
+	 *
+	 * See https://en.wikipedia.org/wiki/Bit_field
+	 *
+	 */
+	flags: ThreadMemberFlags;
+}
+
+/**
+ * Currently undocumented
+ */
```
vladfrangu, line 666: We do not include this for undocumented things, we keep them without a docstring

payloads/v9/channel.ts
```diff
@@ -0,0 +1,1047 @@
...
+	user_id: Snowflake;
+	/**
+	 * An ISO8601 timestamp for when the member last joined
+	 */
+	join_timestamp: string;
+	/**
+	 * Member flags combined as a bitfield
+	 *
+	 * See https://en.wikipedia.org/wiki/Bit_field
+	 *
```
vladfrangu, line 659:
```suggestion
```

rest/v9/oauth2.ts
```diff
@@ -0,0 +1,199 @@
+import type { Permissions, Snowflake } from '../../globals';
+import type { APIApplication, APIGuild, APIUser, APIWebhook, OAuth2Scopes } from '../../payloads/v9/index';
+
+/**
+ * https://discord.com/developers/docs/topics/oauth2#get-current-application-information
+ */
+export type RESTGetAPIOauth2CurrentApplicationResult = Omit<APIApplication, 'flags'>;
```
vladfrangu, line 7: For the sake of my sanity, please rename the `Oauth` to `OAuth` (notice the capital A). I will do this for API v8

vladfrangu (changes requested):

payloads/v9/channel.ts
ckohen: do you have a suggested naming scheme for this?

vladfrangu:
Exactly like the client does, 24Hours, OneWeek, etc. See what fits best

payloads/v9/channel.ts
vladfrangu: See https://github.com/discordjs/discord-api-types/pull/133#issuecomment-846378899 since GH really did a good job placing it in the right spot

ckohen: Okay, done, the client calls 1 day 24 hours, but enum names can't start with a number, so I did OneDay, do you think TwentyFourHours would be better.

rest/v9/oauth2.ts
ckohen: There's a couple places in Changelog and rest/common do you want those changed too?

vladfrangu: I mean I checked, there doesn't seem to be anything in rest/common? Can you link me? Also, ignore the changelog

ckohen: There's 3 here: https://github.com/discordjs/discord-api-types/blob/f0b2766d5b55bd9b8b8ba9c506a868dafcdca568/rest/common.ts#L80-L82

And one a little later, they are capitalized in the dev docs

vladfrangu: Oh fuck. Do those in API v9, I will quickly unfuck that in v8. Nice catch, thanks a bunch <3

vladfrangu: Wait, I'm blind, thought that was an import. Will fix

payloads/v9/channel.ts
vladfrangu: > Okay, done, the client calls 1 day 24 hours, but enum names can't start with a number, so I did OneDay, do you think TwentyFourHours would be better.

Nahhhh, OneDay is fine! TwentyFourHours is just a mouthful

vladfrangu:
Is this still a draft, or can we get this merged?

ckohen:
It's only a draft because the API docs may not be finalized, I'm fine with merging granted it may come at the cost of further PRs

vladfrangu:
I'm gonna merge this as-is for now, as I want to get everything up to date relatively soon. Further changes, please PR them! 💙

vladfrangu (approved):

PR #156, opened 2021-07-18
chore: Get up to date *again*

vladfrangu:
**Please describe the changes this PR makes and why it should be merged:**

~~Draft for now, as I still have lots to do :megu:~~

- Updates dependencies
- Cleans up message buttons, makes them stricter too
- Brings some oauth2 docs up tod ate
- Corrects the key for the `nick` change, it's now ACTUALLY `nick` instead of what it was previously
- Adds the missing `application_id` property to messages
- Fixes the unversioned API bases in rest exports
- Adds all new guild features to v9 (and v8 once @advaith1 and I talk)
- Fixes the rudimentary interactions types, making them stricter and cleaner
- If you're like me and want to test your unions, tsd is now present, so have at it!

**Reference Discord API Docs PRs or commits:**

PR #160, opened 2021-07-20
chore: Add more missing stuff

advaith1:
#156 did not, actually, totally get up to date
hopefully, this does *but no promises*

- fix docs links to application object
- Guild Member Update member joined_at is nullable
- fix docs links to Update Presence
- move application object & flags to new file
- update embed desc limit to 4096
- update guild max_presence desc
- add team.name
- add application webhook type
- add missing json error codes (there were a lot)
- update channel name min char limit to 1
- document sending/editing components in messages
- embeds is nullable while editing
- add missing "see" links to edit msg
- add adding attachments while editing
- add removing attachments while editing
- add missing "see" links to execute webhook
- add application-owned webhook restriction for components
- add editing components in webhook messages
- add removing attachments in webhook messages
- moves execute webhook thread_id from json body to query


gateway/v8.ts
```diff
@@ -1624,3 +1625,7 @@ interface MessageReactionRemoveData {
 	guild_id?: Snowflake;
 }
 // #endregion Shared
+
+type Nullable<T> = {
```
kyranet, line 1629: I'd rename this to `NullableProperties`, and maybe move it to a common place, seeing this type is duplicated in the v9 file as well.

vladfrangu: Can go in utils/internals.ts, and never get exported publicly

vladfrangu (changes requested):

gateway/v8.ts
advaith1: the `Nullable` name was already used before and is consistent with the built-in `Required` and `Optional`, but I've moved it

vladfrangu (approved):

kyranet (approved):

PR #162, opened 2021-07-20
feat(Interactions): add interaction response and followup route

advaith1:
other than Create Interaction Response, these are just aliases for the equivalent webhook routes, but not including them is inconsistent and can be confusing to the user

this also moves a Get Webhook Message type up so it isn't sandwiched between Edit Webhook Message types anymore

### DIFF
--- a/rest/v9/interactions.ts
+++ b/rest/v9/interactions.ts
@@ -3,7 +3,16 @@ import type {
 	APIApplicationCommandPermission,
 	APIGuildApplicationCommandPermissions,
 	APIInteractionResponse,
+	APIInteractionResponseCallbackData,
 } from '../../payloads/v9/index';
+import type {
+	RESTDeleteAPIWebhookWithTokenMessageResult,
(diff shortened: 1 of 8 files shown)


rest/v9/interactions.ts
```diff
@@ -111,6 +120,87 @@ export type RESTPostAPIInteractionCallbackFormDataBody =
 			file: unknown;
 	  });
 
+/**
+ * https://discord.com/developers/docs/interactions/slash-commands#get-original-interaction-response
+ */
+export type RESTGetAPIInteractionOriginalResponseResult = RESTGetAPIWebhookWithTokenMessageResult;
```
vladfrangu, line 126: Any reason these aren't in v8 too?

vladfrangu (changes requested):

rest/v9/interactions.ts
advaith1: oops

vladfrangu (approved):

kyranet (approved):

PR #167, opened 2021-07-25
feat: thread updates

advaith1:
* https://github.com/discord/discord-api-docs/pull/3502

### DIFF
--- a/rest/v9/guild.ts
+++ b/rest/v9/guild.ts
@@ -12,6 +12,7 @@ import type {
 	APIGuildWidget,
 	APIGuildWidgetSettings,
 	APIRole,
+	APIThreadList,
 	APIVoiceRegion,
 	GuildDefaultMessageNotifications,
 	GuildExplicitContentFilter,
@@ -302,6 +303,11 @@ export type RESTPatchAPIGuildChannelPositionsJSONBody = Array<{
  * https://discord.com/developers/docs/resources/guild#modify-guild-channel-positions
  */
 export type RESTPatchAPIGuildChannelPositionsResult = never;
+
+/**
+ * https://discord.com/developers/docs/resources/guild#list-active-threads
+ */
+export type RESTGetAPIGuildThreadsResult = Omit<APIThreadList, 'has_more'>;
 
 /**
  * https://discord.com/developers/docs/resources/guild#get-guild-member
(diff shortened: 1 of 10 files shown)


rest/v9/guild.ts
```diff
@@ -303,6 +304,11 @@ export type RESTPatchAPIGuildChannelPositionsJSONBody = Array<{
  */
 export type RESTPatchAPIGuildChannelPositionsResult = never;
 
+/**
+ * https://discord.com/developers/docs/resources/guild#list-active-threads
+ */
+export type RESTGetAPIGuildThreadsResult = Omit<APIThreadList, 'has_more'>;
```
vladfrangu, line 310: Can't you just remove the has_more since it's getting yeeted anyways?

vladfrangu, line 310: Just a question since phone review, is this route in the routes object?

vladfrangu (changes requested):

rest/v9/guild.ts
advaith1: the get channel (public/private/joined private) archived threads routes are keeping it, it's only removed from get guild active threads

advaith1: oops, was gonna do it but forgot, will add

rest/v9/index.ts
```diff
@@ -435,9 +435,19 @@ export const Routes = {
...
+	 * - GET `/guilds/{guild.id}/threads/active`
+	 */
+	guildActiveThreads(guildId: Snowflake) {
+		return `/guilds/${guildId}/threads/active` as const;
+	},
+
 	/**
 	 * Route for:
 	 * - GET `/channels/{channel.id}/threads/active`
+	 * 	 (deprecated, use [List Active Guild Threads](https://discord.com/developers/docs/resources/guild#list-active-threads) instead.
```
vladfrangu, line 449: Use `@deprecated` instead

vladfrangu (changes requested):

rest/v9/index.ts
advaith1: didnt seem to render inline, and that might mark the whole function as deprecated while its only deprecated for that particular route

vladfrangu (approved):

PR #169, opened 2021-07-26
fix: change resolved index types to string

advaith1:
this lets you access resolved fields using ts 4.3.

this should be reverted once ts 4.4 releases ([august 24](https://github.com/microsoft/TypeScript/issues/44237))


vladfrangu (approved):
Thanks TypeScript.

PR #179, opened 2021-08-08
fix(MessageComponent): emoji.name is not optional

Dougley:

`[APIMessageComponentEmoji].name` is always required.

https://discord.com/developers/docs/resources/emoji#emoji-object-emoji-structure
> name: ?string **(can be null only in reaction emoji objects)**

### DIFF
--- a/deno/payloads/v8/channel.ts
+++ b/deno/payloads/v8/channel.ts
@@ -970,7 +970,7 @@ export interface APIMessageComponentEmoji {
 	/**
 	 * Emoji name
 	 */
-	name?: string;
+	name: string;
 	/**
 	 * Whether this emoji is animated
 	 */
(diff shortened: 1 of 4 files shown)


vladfrangu:
We've tested it, and it's optional for custom emojis. Unless that changed recently.. I don't think this is valid 👀

Dougley:
Interesting, the docs seem to suggest it's required due to being supplied in all the examples and the docs stating it's only null for reactions. Incoming interactions will always have this field filled however, so ¯\\\_(ツ)_/¯

vladfrangu:
Oh, greatttt... its a send/receive difference... Can you change the one in POST then to be the Partial of the interface you modified?

Dougley:
Since components are part of action rows it's impossible to just `Partial` this easily, I think that changing how the types are referenced would overextend past the scope for this PR. I'm happy to close this and just patch around this in my own code instead 😄

vladfrangu:
Hmm... yeah for REST you'd have to change it entirely... :notlikethis:...

In your code you can just slap a `!` after `.name` or whatever property you need. I'd say keep the PR open for now, I can do it later if you'd like :D

PR #185, opened 2021-08-22
feat(threads): add `invitable`

advaith1:
* https://github.com/discord/discord-api-docs/pull/3600

### DIFF
--- a/rest/v9/channel.ts
+++ b/rest/v9/channel.ts
@@ -130,6 +130,12 @@ export interface RESTPatchAPIChannelJSONBody {
 	 * Channel types: text, news
 	 */
 	default_auto_archive_duration?: ThreadAutoArchiveDuration;
+	/**
+	 * Whether non-moderators can add other non-moderators to the thread
+	 *
+	 * Channel types: privateThread
+	 */
+	invitable?: boolean;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


rest/v9/channel.ts
```diff
@@ -130,6 +130,12 @@ export interface RESTPatchAPIChannelJSONBody {
 	 * Channel types: text, news
 	 */
 	default_auto_archive_duration?: ThreadAutoArchiveDuration;
+	/**
+	 * Whether non-moderators can add other non-moderators to the thread
+	 *
+	 * Channel types: privateThread
```
vladfrangu, line 136: What's this?

vladfrangu (changes requested):

rest/v9/channel.ts
advaith1: ?

vladfrangu: What does it represent? Can you be more specific wkth the jsdoc, like `Only present on the following channel typez: {@link ChannelTypes.PrivateThread}` or similar

advaith1: well all the other fields use the same format, should I change all of them? that feels out of the scope of this PR

vladfrangu: Do they..? Ugh.. Yeah, I guess for consistency, leave it as is for now, but we gotta fix that

vladfrangu (approved):

PR #205, opened 2021-09-23
feat: add autocomplete api types

suneettipirneni:

Adds *very experimental* support for autocomplete.

- https://github.com/discord/discord-api-docs/pull/3849


rest/v9/webhook.ts
```diff
@@ -135,6 +136,10 @@ export interface RESTPostAPIWebhookWithTokenJSONBody {
 	 * See https://discord.com/developers/docs/interactions/message-components#component-object
 	 */
 	components?: APIActionRowComponent[];
+	/**
+	 * Autocomplete choices (limited to 25 choices)
+	 */
+	choices?: APIApplicationCommandOptionChoice[];
```
vladfrangu, line 142: This is so not where this belongs

payloads/v9/interactions.ts
```diff
@@ -19,7 +20,11 @@ export * from './_interactions/applicationCommands';
 /**
  * https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object
  */
-export type APIInteraction = APIPingInteraction | APIApplicationCommandInteraction | APIMessageComponentInteraction;
+export type APIInteraction =
+	| APIPingInteraction
+	| APIApplicationCommandInteraction
+	| APIMessageComponentInteraction
+	| APICommandAutocompleteInteraction;
```
vladfrangu, line 27: This should be named APIApplicationCommandAutocompleteInteraction

payloads/v9/_interactions/responses.ts
```diff
@@ -24,6 +25,10 @@ export interface APIInteractionResponsePong {
 	type: InteractionResponseType.Pong;
 }
 
+export interface APICommandAutocompleteResponse {
+	type: InteractionResponseType.ApplicationCommandAutocompleteResult;
```
vladfrangu, line 29: Where's the data

payloads/v9/_interactions/_applicationCommands/chatInput.ts
```diff
@@ -40,18 +40,29 @@ export interface APIApplicationCommandSubCommandOptions extends Omit<APIApplicat
...
  * but they can have a `choices` one
  */
-export interface APIApplicationCommandArgumentOptions extends Omit<APIApplicationCommandOptionBase, 'type'> {
+export type APIApplicationCommandArgumentOptions = Omit<APIApplicationCommandOptionBase, 'type'> & {
 	type:
 		| ApplicationCommandOptionType.String
 		| ApplicationCommandOptionType.Integer
 		| ApplicationCommandOptionType.Number;
-	choices?: APIApplicationCommandOptionChoice[];
+} & AutoCompleteOrChoices;
```
vladfrangu, line 60: I don't think this is valid... doesn't it also work for numbers?

payloads/v9/_interactions/_applicationCommands/chatInput.ts
```diff
@@ -40,18 +40,29 @@ export interface APIApplicationCommandSubCommandOptions extends Omit<APIApplicat
...
+
+export interface APIApplicationCommandAutocompleteOptions
+	extends Omit<APIApplicationCommandArgumentOptions, 'choices' | 'autocomplete'> {
+	choices?: never;
```
vladfrangu, line 64: Don't include this

payloads/v9/_interactions/_applicationCommands/chatInput.ts
```diff
@@ -40,18 +40,29 @@ export interface APIApplicationCommandSubCommandOptions extends Omit<APIApplicat
 	options?: APIApplicationCommandOption[];
 }
 
+// This is here because auto completed command options aren't a seperate 'type', and it only
+// applies to the the choices types.
+type AutoCompleteOrChoices =
+	| { type: ApplicationCommandOptionType.String; choices?: never; autocomplete: true }
```
vladfrangu, line 46: Don't include choices in this case

vladfrangu (changes requested):
Initial review - don't forget v8 :D

suneettipirneni:
So unfortunately typescript won't be able to enforce that the command has either a `choices` array OR a `autocomplete` boolean, without breaking the existing union, so it's just going to be appended onto the `APIApplicationCommandArgumentOptions`, and will need to be enforced during runtime.

ckohen:
When *sending only* typescript can enforce either choices or autocomplete via a discriminated union. On the receiving side there's not much of a benefit to having a discriminated union, but it could still work, just means people have to build type guards

suneettipirneni:
> When _sending only_ typescript can enforce either choices or autocomplete via a discriminated union. On the receiving side there's not much of a benefit to having a discriminated union, but it could still work, just means people have to build type guards

Even on the sending side with a discriminated union it doesn't work properly, it's because TS unions aren't 'exclusive or' unions. That means if multiple types with the same tag (`type`) in this case are used, the discriminated union will no longer function properly because there's no longer a 1-1 mapping of tags to types.

suneettipirneni:
Here's a TS playground that shows this: 
https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMoEcCucooN4BQyxyYAngA4QBcyA5AM5Y4R0DcRJANhCAOZgAFrRCYAtgCNoHAL4EC5KsgDCwKAh7IAvMkLFFNZACIEajRCMAaTlDgATYJgYjxUqNbkFQkWIhQBBEAB7IWhVdU09EgNaOlMI1msSZDhg0KgAJXtHZ2RRSWkCTwM0QTglHXDzZAAfZECQwTCzTTqMbFwOAgQgkAYwZDs4MDhaVDKK3U59SkM4lsTplLSmzOynWgBGAAZtpJJbBw3kHb3kAHpz5AAhRuRDnOW7ZcboLKOGFNwUri4ggHcIM9VigpAg4E4UAAVVDITAgYC9T4sEB0AZiTBgbC-MjICAADw0TmAADcIAA6IpAA

ckohen:
You need a discriminated union explicitly preventing the other property. But I realized there's no identifier for which we would be using without adding a nonexistent prop. How does discord handle it if you send both anyways? Maybe a type generic could work?

suneettipirneni:
> You need a discriminated union explicitly preventing the other property. But I realized there's no identifier for which we would be using without adding a nonexistent prop. How does discord handle it if you send both anyways? Maybe a type generic could work?

Maybe I'm missing something, but I don't know how a generic type would help here, since generics don't change which fields are shown and which ones are not shown. Generics only affect already existing fields. As for how discord handles it I'm not entirely sure since I haven't tested it, but according the docs PR, `choices` can't be present when `autocomplete` is true.

ckohen:
[This}(https://www.typescriptlang.org/play?#code/FASwdgLgpgTgZgQwMZQAQHkAOEQHswDqIEAFgMIm4goA8AggK4S6pQAe0YAJgM6oBGuXABsoCMKgC8qRMJ5QAfKgDewVOtQQAnpigAuVAGZUAH1QAWANxqNYBAFt9qHhBjgA5tY2ouUHkjdsPDADFzcwTxt1JEpqPwB+A0ZmVg4obj5XBjR41DAoADdYVAMAbQBdLw0EJlwkXHtMUWhE1GSWdk5eGQQ5HLzC4oNBETEJMwZuKDhwKC5rAF9gYHqwFx8ECAQDLBx8IlIKKlosxSkVKM0dJ0MAGku7RwMAchrmesbmqGf7719-QJ7EKoV4+PwBEBBfA-S5vOoNJpQaAGU73BZAA) is what I mean by using generics

suneettipirneni:
> [This](https://www.typescriptlang.org/play?#code/FASwdgLgpgTgZgQwMZQAQHkAOEQHswDqIEAFgMIm4goA8AggK4S6pQAe0YAJgM6oBGuXABsoCMKgC8qRMJ5QAfKgDewVOtQQAnpigAuVAGZUAH1QAWANxqNYBAFt9qHhBjgA5tY2ouUHkjdsPDADFzcwTxt1JEpqPwB+A0ZmVg4obj5XBjR41DAoADdYVAMAbQBdLw0EJlwkXHtMUWhE1GSWdk5eGQQ5HLzC4oNBETEJMwZuKDhwKC5rAF9gYHqwFx8ECAQDLBx8IlIKKlosxSkVKM0dJ0MAGku7RwMAchrmesbmqGf7719-QJ7EKoV4+PwBEBBfA-S5vOoNJpQaAGU73BZAA) is what I mean by using generics

I see how this can work, however, how would you integrate this into the `APIApplicationCommandOption` type union?

suneettipirneni:
Alright so, somehow I was able to get typescript to typecheck the `autocomplete` variants. I basically made the `autocomplete` property of type `never` on all types except on choice types. When choices are provided, autocomplete is defined as false or undefined. If `autocomplete` is present `choices` is defined as `never`. This results in invalid variants not being allowed by typescript.

payloads/v8/_interactions/_applicationCommands/chatInput.ts
```diff
@@ -20,6 +20,7 @@ interface APIApplicationCommandOptionBase {
 	description: string;
 	default?: boolean;
 	required?: boolean;
+	autocomplete?: never;
```
vladfrangu, line 23: Why are you defining it here as optionally never o.o

payloads/v8/_interactions/autocomplete.ts
```diff
@@ -0,0 +1,25 @@
+import type { APIBaseInteraction, ApplicationCommandOptionType, InteractionType } from '..';
```
vladfrangu, line 1: Please fix this import so it points directly to the file rather than to `..` (for instance `../index` or w/e)

payloads/v8/_interactions/autocomplete.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type APIApplicationCommandAutocompleteInteraction = APIBaseInteraction<
+	InteractionType.ApplicationCommandAutocomplete,
+	APICommandAutocompleteInteractionData
+>;
+
+export interface APICommandAutocompleteInteractionData {
```
vladfrangu, line 8: This is not at all what data you receive from this interaction event 🤔
Check the notion document posted by devsnek in Discord Developer's (stage-info channel)

payloads/v8/_interactions/responses.ts
```diff
@@ -76,3 +87,7 @@ export type APIInteractionResponseCallbackData = Omit<
 	RESTPostAPIWebhookWithTokenJSONBody,
 	'username' | 'avatar_url'
 > & { flags?: MessageFlags };
+
+export interface APICommandAutocompleteInteractionResponseCallbackData extends APIInteractionResponseCallbackData {
```
vladfrangu, line 91: No..thats also not the interface / response this type expects... It only expects the choices array 😓

rest/v8/interactions.ts
```diff
@@ -35,6 +35,10 @@ type RESTPostAPIBaseApplicationCommandsJSONBody = Omit<
  */
 export interface RESTPostAPIChatInputApplicationCommandsJSONBody extends RESTPostAPIBaseApplicationCommandsJSONBody {
 	type?: ApplicationCommandType.ChatInput;
+	/**
+	 * Whether this application command option should be autocompleted
+	 */
+	auto_complete?: boolean;
```
vladfrangu, line 41: Why is this separated here??

vladfrangu (changes requested):
More issues, please also port the fixes to v9 :D

payloads/v8/_interactions/_applicationCommands/chatInput.ts
suneettipirneni: It's weird, but this is what was I was told to do in the TS discord in order to get typechecking to work properly for the `autocomplete` field, while also preserving the union. I'm pretty sure this is the only way this can work.

PR #250, opened 2021-12-01
types: make application command option union easier to use

suneettipirneni:

The previous way that autocomplete was enforced create types that were hard/near impossible to use. This PR sacrifices the type strength for usability making it waaay easier to adopt. So essentially `choices` and `autocomplete` are no longer mutually exclusive.

### DIFF
--- a/deno/payloads/v8/_interactions/_applicationCommands/chatInput.ts
+++ b/deno/payloads/v8/_interactions/_applicationCommands/chatInput.ts
@@ -25,11 +25,14 @@ interface APIApplicationCommandOptionBase {
  * https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-structure
  */
 export type APIApplicationCommandOption =
-	| APIApplicationCommandChoiceArgumentOptions
+	| APIApplicationCommandStringArgumentOptions
 	| APIApplicationCommandSubCommandOptions
+	| APIApplicationCommandOptionBase
 	| APIApplicationCommandChannelOptions
 	| APIApplicationCommandOptionBase
-	| APIApplicationCommandNumberArgumentOptions;
+	| APIApplicationCommandNumberArgumentOptions
+	| APIApplicationCommandStringAutocompleteOptions
+	| APIApplicationCommandNumericAutocompleteOptions;
 
 /**
  * This type is exported as a way to make it stricter for you when you're writing your commands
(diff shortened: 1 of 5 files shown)


vladfrangu:
I don't understand how this fixes anything. If anything, you should've added `autocomplete?: false` to the options that can also have choices, instead of merging them again in one type

suneettipirneni:
> I don't understand how this fixes anything. If anything, you should've added `autocomplete?: false` to the options that can also have choices, instead of merging them again in one type

That won't work because of the way discriminated unions work.

In our case we use `type` as the tag for each union. If we split the autocomplete options up and the choices options up, we'll end up with two interfaces that both have `type` set to string number or integer. 

Because there's two interfaces with the same `type` values. It breaks the union and it'll no longer work properly.

vladfrangu:
No it won't? [Playground Link](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgIIFcwHsFYLYAOANhGHAEYkDKYUoA5sgN4BQy7yYAngRAFzIA5AGdaDQQG42HOJhz5ipfpyjoIUgL4sWoSLEQoAclhAZsuQiTKUINOiEasOnHspFiHk6e1nmFViAB+AXgiYXVvZAQACyxgJGFg5FF7egBtAF1NbW5eZABVEGATZABeNDkLRWtqD0YAH2RjU0r-Ugpa1KkWXBBRZCxyACsBQuKQMuZI3LcU8QAaSN95SyUBWjVF7W3nGLiEgUyWDSA)

image

Are you sure this is breaking anything? Could you build a playground url where it breaks?

suneettipirneni:
> Easy fix. [Playground Link](https://www.typescriptlang.org/play?#code/KYOwrgtgBA8gDgFwJYHsQBUCedhQN4BQUxUAyggE5IgDmANESQHKQBGwFDJUAkiAsBocGAXwIFqAigDMAhgGNcLCOwrxkafI2IJswAFyxEqDHoB0y1QG5tUeQAsUSRQGdDskJgDaAXRvdZMAQUeRQIOAAbYAEAfkM5CJdgGzEJfg45RTJKahp1Ey1uXRxDfLQsHDNyKlp-EgcnV3dPXzriQODQ8KjY+NlE5IJUyQyFXD4BITVjTUIivVKZ00qJwQ42qA6QsMjo4DioBKSU8RGZMagAQSCUatyykEKSYoMjDWXgKpza2y2u3YEhkoYEGw3S5yy12CqymDyeOgWbxMFU+MPWvxu-x6r2BoNO4MyuChKEsHDhc2eiIeKIsbHRAUxO2xQIoIJOBBeSM0AF5eOlYUsoAAfKCk6bvYXZGp5QUi4l3WhwuU3NHigrK4Jih42AgAel1UHQ9iQLjsYVY1FkEpNIAA5AgoHAUC4XEhWFE6FBMNFDaRNhEIigAO6mpAIMwEUIgFwOlBLRYS3kUhElLkfMyqrj1RzOYBuKC+T0YzpMvYskGiIA)

🤦‍♂️ I'm dumb

vladfrangu:
Are you sure that fixes it? because it also looks wrong

kyranet:
When in doubt, add tests, lots of type tests. That should tell us whether or not they're fixed.

kyranet:
Oh, I didn't see that directory.

> How do you unit test... types?

Just like how we have been unit testing types in the main library for months? Even this project's tests test types:

https://github.com/discordjs/discord-api-types/blob/c7efcd55059673ab9fc8e6ef9711050700274057/tests/v9/interactions.test-d.ts#L26-L27

kyranet (approved):

vladfrangu (approved):

PR #258, opened 2021-12-08
fix(BasePayload): `s` and `t` can be nullable

NurMarvin:
**Please describe the changes this PR makes and why it should be merged:**
In some cases the `s` and `t` fields on gateway payloads can be `null`, e.g. in case of op code `10` (Hello). This was previously not reflected in the typings, which this PR changes.

**Reference Discord API Docs PRs or commits:**
https://github.com/discord/discord-api-docs/blob/master/docs/topics/Gateway.md?plain=1#L28-L31

PR #259, opened 2021-12-08
fix(NonDispatchPayload): `t` & `s` fields are always null on non-dispatch payloads

NurMarvin:
This PR does pretty much does the same thing as #258, except in a more elegant way. #258 would have made the field `t` and `s` always nullable, not just on non-dispatch payloads like this PR now does.

https://github.com/discord/discord-api-docs/blob/master/docs/topics/Gateway.md?plain=1#L28-L31

### DIFF
--- a/deno/gateway/v8.ts
+++ b/deno/gateway/v8.ts
@@ -1568,7 +1568,10 @@ interface BasePayload {
 	t?: string;
 }
 
-type NonDispatchPayload = Omit<BasePayload, 't'>;
+type NonDispatchPayload = Omit<BasePayload, 't' | 's'> & {
+	t: null;
+	s: null;
+};
 
 interface DataPayload<Event extends GatewayDispatchEvents, D = unknown> extends BasePayload {
 	op: GatewayOpcodes.Dispatch;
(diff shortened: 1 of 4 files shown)


SpaceEEC (approved):

kyranet (commented):
Does this not happen in v8? If it does, it should be updated accordingly.

Also, the Deno scripts need to be run, I don't see changes to them in this PR.

NurMarvin:
I just checked and this happens in all available gateway versions (6, 7, 8 and 9), so I will be adding these changes to the v8 and v6 files as well. Running the Deno script wasn't mentioned anywhere, so I wasn't aware that I had to run it but I will be also doing that as well.

vladfrangu:
V6 is not getting updates anymore

NurMarvin:
> 
> As for the deno script, if you installed node deps it should've automatically do it

Ah, my bad. I did install the node deps but made my commit using a GUI client which seems to have bypassed the pre-commit hook, so the deno script was not executed.

kyranet (approved):

vladfrangu (approved):

PR #263, opened 2021-12-09
fix: only a partial object is needed when updating attachments

cyyynthia:
This PR changes the type of the `attachments` field of every message update body from a complete attachment object to only a partial object, with only the ID required.

It is currently impossible to send a query with all the information the typings are asking for, and sending fields like `filename` for already uploaded files results in an error from the API.

None


vladfrangu (approved):

vladfrangu (approved):

vladfrangu:
Woops. Lol

PR #282, opened 2021-12-29
fix(Snowflake): Stricter typing

Syjalo:
Minor `Snowflake` typing changes to make it more stricter


vladfrangu:
We've done this before but it was reverted due to complaints 😓

PR #283, opened 2021-12-29
fix(package.json) missing main attribute

Flam3rboy:
I'm using ``discord-api-types`` in a react-native project, however bundling fails because ``package.json`` doesn't have a ``main`` attribute, which indicates the main entry point of the module.
That's why I set ``"main"`` to the current up to date file ``"./v9.js"``.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
 	"version": "0.26.0",
 	"description": "Discord API typings that are kept up to date for use in bot library creation.",
 	"types": "./v9.d.ts",
+	"main": "./v9.js",
 	"exports": {
 		"./globals": {
 			"require": "./globals.js",


suneettipirneni:
The error is resolved by versioning your import `discord-api-types/v9`. It's intentional that `main` isn't defined in the package.json because when a new api version comes out it will break a lot of code for users using the library.

vladfrangu:
This was an intentional removal (did I not document it in the changelog..possibly)

As @suneettipirneni said, you're expected to import from the version you're targetting (at this time v6, v8 or v9, and v4 for voice), which also should solve your build error!
If that's not the case, let us know in an issue (preferably with a repro sample)!

PR #287, opened 2022-01-05
chore: bump LICENSE year

GodderE2D:
Another year, another bump. Bumps the license year to 2022!

Note: "2020 - 2022" instead for consistency.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2020 vladfrangu
+Copyright (c) 2020 - 2022 vladfrangu
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


vladfrangu:
Thank you for the PR, however I don't see a reason to append a year range to the license. As such, I'll be closing this PR, but if you think adding a year range is beneficial, please feel free to comment here and I'll reopen the PR. 😁

GodderE2D:
@vladfrangu due to other discord.js repositories with the MIT license, some of them have a year range. However, now thinking about it, it might not be as beneficial as much and keep them the same for now.

I could just remove the year range and just add 2022 if you would like to.

vladfrangu:
I wouldn't want to change the year at all, a short explanation as to why can be seen here: https://opensource.stackexchange.com/a/1523

Specifically this part explains it:
>`[year]` is the **year copyright was applied**. If you apply the copyright (which happens automatically when you write the work) in 2015, put in 2015. Some people like to put in a range of years (i.e. 2015-2020) - this isn't necessary.

PR #292, opened 2022-01-09
feat(Routes): add route for oauth2 application assets

vaporox:

Adds a new route for application assets. I'm actually not sure where this is documented but discord.js uses it so it should exist.

https://github.com/discordjs/discord.js/blob/main/packages/discord.js/src/structures/interfaces/Application.js#L108


vladfrangu:
Please also add the respective results from the API calls

PR #306, opened 2022-01-26
refactor: change const enums to regular enums

### LINKED ISSUE #295: Move from const enums
**Is your feature request related to a problem? Please describe.**
I need to access one of the enums (`ApplicationCommandOptionType`) but do it dynamically, currently it's a const enum so you are unable to do that

**Describe the ideal solution**
It's made a non const enum

**Describe alternatives you've considered**
Copying and pasting the enum into my project (I will then have to make sure I constantly update this which works but isn't ideal)

**Additional context**
Would be worth doing this on discord.js itself if there aren't plans to fully use this package on djs

suneettipirneni:

Changes all regular enums in v8 and v9 to const enums. `PermissionFlagBits` is still an object since it contains `bigint` values. The other `*Flags` enums are also still objects because the docs state they are still of the `integer` data type.

Closes #295 

### DIFF
--- a/deno/voice/v4.ts
+++ b/deno/voice/v4.ts
@@ -3,7 +3,7 @@ export const VoiceGatewayVersion = '4';
 /**
  * https://discord.com/developers/docs/topics/opcodes-and-status-codes#voice-voice-opcodes
  */
-export const enum VoiceOpcodes {
+export enum VoiceOpcodes {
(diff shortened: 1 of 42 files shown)


vladfrangu (approved):

ghostdevv:
Thank you!

PR #319, opened 2022-02-10
chore: cleanup and improve CI

vladfrangu:

- @ next deploys happen after the latest commit to master (faster releases woo)
- automatic PRs for new releases every Tuesday and Thursday if needed
- deno types are now not going to try to autogenerate and push but rather warn the user (since it can't push to forks anyways)
- Labels are getting changed to make them more useful
- Deps updated


tsconfig.eslint.json
```diff
@@ -15,7 +14,7 @@
 		"rpc/**/*.ts",
 		"utils/**/*.ts",
 		"voice/**/*.ts",
-		"scripts",
+		"scripts/**/*",
```
favna, line 17: This is the same. Globstars are implies in tsconfig.

scripts/actions/report-deno-not-ran.ts
```diff
@@ -0,0 +1,87 @@
+import { Octokit } from '@octokit/action';
```
favna, line 1: If you really want TS then fine but IMHO you're adding a LOT of overhead (`ts-node`, extra tsconfig file, extra code in `tsconfig.eslint.json`, extra script in package.json, not having TLA) by having that over a `.mjs` file.

Personally I used to do TS based scripts too... 2 years ago. But ESM is just so much easier. Especially now that you've written (and presumably tested) it, you don't *really* need the type information anymore.

scripts/actions/report-deno-not-ran.ts
```diff
@@ -0,0 +1,87 @@
...
+import type { PullRequestOpenedEvent, PullRequestSynchronizeEvent } from '@octokit/webhooks-types';
+import { execSync } from 'child_process';
+import { readFile } from 'fs/promises';
```
favna, line 4:
```suggestion
import { execSync } from 'node:child_process';
import { readFile } from 'node:fs/promises';
```

package.json
```diff
@@ -110,29 +111,32 @@
 		"node": ">=12"
 	},
 	"devDependencies": {
-		"@babel/core": "^7.16.5",
-		"@babel/eslint-parser": "^7.16.5",
+		"@babel/core": "^7.17.2",
+		"@babel/eslint-parser": "^7.17.0",
 		"@babel/plugin-syntax-top-level-await": "^7.14.5",
```
favna, line 116: Are these deps still required? You *at least* removed `@babel/eslint-parser` from the eslint config below.

.github/workflows/publish-next.yml
```diff
@@ -2,14 +2,23 @@ name: Publish next
...
 jobs:
   publish:
-    name: Publish to npm
+    name: Publish @next release to npm
     runs-on: ubuntu-latest
     steps:
+      - name: Cancel previous publish attempts
+        uses: styfle/cancel-workflow-action@0.9.1
+        with:
+          access_token: ${{ github.token }}
```
favna, line 20: Y'know... seeing this here is really kinda ironic after you said just before .. well this 

image

IMHO you shouldn't have this here because there are perfectly valid reasons for publishing both `0.x.y-asd822a` and `0.x.yrty32a` even though those 2 commits were added right after one another, the second while CI of the first was still running.

favna (changes requested):
You should add an auto deprecate workflow for `@favware/npm-deperecate`  to auto deprecate `next` versions. And before you say "the publish-next workflow already does that", the main DJS repo also uses it these days: https://github.com/discordjs/discord.js/blob/main/.github/workflows/auto-deprecate.yml

---

Also if you ask me this should go yarn v3 to match the DJS repo so the same package manager is used org-wide but that'll probably be rejected by Vladdy coz Vladdy is Vladdy and Vladdy likes long things, both names and running processes. ¯\\\_(ツ)_/¯

scripts/actions/report-deno-not-ran.ts
vladfrangu: I couldn't get ts-node to properly make the script it runs run as ESM (because node reads from package.json and well..yknow)

Having TLA is really not something I crave thaaat much, and the small extras which I can probably remove now. I'll wait for thoughts from the rest, but imo I'd say it's fine, especially since its fast nuff.

.github/workflows/publish-next.yml
vladfrangu: More concerned over ratelimits / deprecate cycles clashing, so I'd say it's fine to have it now

package.json
vladfrangu: Still having an interesting experience with mjs and eslint sadly, but I might not need this anymore

favna: if you use @typescript-eslint/parser you need to define `extraFileExtensions: ['.mjs']`. That might just be all you need. That's all that is done for `.mjs` support for Sapphire anyway and we have maaany mjs scripts.

See: https://github.com/sapphiredev/utilities/blob/9e8ac5661e9a05299cba5bd58eb69d435676359f/packages/eslint-config/src/index.ts#L14

.eslintrc.json
```diff
@@ -0,0 +1,11 @@
+{
+	"extends": "marine/prettier/node",
+	"parser": "@typescript-eslint/parser",
+	"parserOptions": {
+		"project": "./tsconfig.eslint.json",
```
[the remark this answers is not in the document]
vladfrangu: It's not, the default points to tsconfig.json, which has far less files than the eslint one 👍

package.json
```diff
@@ -106,62 +107,37 @@
...
 		"is-ci": "^3.0.1",
-		"lint-staged": "^12.1.4",
+		"lint-staged": "^12.3.3",
 		"npm-run-all": "^4.1.5",
 		"prettier": "^2.5.1",
 		"pretty-quick": "^3.1.3",
 		"rimraf": "^3.0.2",
-		"tsd": "^0.19.0",
-		"typescript": "^4.5.4"
+		"ts-node": "^10.5.0",
```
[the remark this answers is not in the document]
vladfrangu: Nop. Forgot to remove it

.github/workflows/labelsync.yml
```diff
@@ -0,0 +1,22 @@
+name: Label Sync
+on:
+  schedule:
+    - cron: '0 0 * * *'
```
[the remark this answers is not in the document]
vladfrangu: In the unlikely event Actions have a choke while we update labels, it doesn't hurt to have a cron too 👍

.github/workflows/create-pr-for-release-and-publish.yml
```diff
@@ -0,0 +1,90 @@
+name: Create PR for new releases and publish new version
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: '0 12 * * 2,4'
```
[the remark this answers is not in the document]
vladfrangu: Just for you, a cron-splanation https://github.com/discordjs/discord-api-types/commit/0f140af20ff949720eab731e53965759c09f5bdb

favna (changes requested):

.github/workflows/check-deno.yml
```diff
@@ -0,0 +1,28 @@
+name: Deno
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
```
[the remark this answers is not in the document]
vladfrangu: I much prefer being explicit, especially if GitHub ever changes the defaults

.github/workflows/create-pr-for-release-and-publish.yml
```diff
@@ -0,0 +1,91 @@
...
+      - main
+  schedule:
+    # Run every Tuesday and Thursday at 12:00
+    - cron: '0 12 * * 2,4'
+
+jobs:
+  create_pr:
+    name: Create PR for new version
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'discordjs' && github.event_name != 'push'
```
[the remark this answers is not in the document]
vladfrangu: Yep, that's the plan, so in the event we need to do a manual deploy we can

favna (approved):

.github/workflows/check-deno.yml
```diff
@@ -0,0 +1,28 @@
+name: Deno
+
+on:
```
ckohen, line 3: The script doesn't currently support it, but would it make sense to run this on push to? (that is, if my other comment doesn't make sense,

.github/workflows/cicd.yml
```diff
@@ -28,40 +28,6 @@ jobs:
...
-        run: |
-          git config user.name "github-actions[bot]"
-          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
-
-      - name: Commit and Push New Code
-        run: |
-          git add --all .
-          git commit -m "build: deno build for ${GITHUB_SHA}" || true
-          git push || true
-
```
ckohen, line 64: I say keep this, only run on push of non-main branches. As far as I'm aware that should allow auto building deno on forks unless the fork owner has disabled acions, at which point CI will fail on the pull request.

scripts/actions/report-deno-not-ran.mjs
```diff
@@ -0,0 +1,83 @@
...
+		});
+	}
+
+	process.exit(0);
+}
+
+console.debug(`⚠️ Found ${diff.length} different files.`, { diff });
+
+const messageBody = [
+	`Hey @${login}! Taking a look at commit ${commitUrl}, it seems that you forgot to run \`npm run build:deno\`. Here are the files that are different from the base:`,
```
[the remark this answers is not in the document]
ckohen: doesn't the create release PR run deno scripts so they are up to date for release tags? 👀

ckohen (commented):
Mostly looks good, couple questions though

.github/workflows/cicd.yml
vladfrangu: It hasn't worked before on several PRs as it'd run on our repository with their origin and the token wouldn't permit access

scripts/actions/report-deno-not-ran.mjs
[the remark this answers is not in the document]
vladfrangu: I mean... It does, but people can use deno types from main via url so...

.github/workflows/cicd.yml
ckohen: because it was running on the pull_request event which doesn't have access to the fork, when pushing to the fork, the fork has access to push to itself

scripts/actions/report-deno-not-ran.mjs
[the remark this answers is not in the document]
ckohen: thats what I thought, it sounded like you literally meant only releases

.github/workflows/cicd.yml
vladfrangu: It was running on push and pull_request, and neither worked 🥲

vladfrangu: Oh now I remember, there's two issues with that:
1. If it somehow lacks access to commit the user has no way of knowing
2. If it does have access to commit, the user needs to push an empty commit so we can run CI on it and merge it

Both of these are quite big issues imo, and a forced requested changes till its fixed vs me or whomever is reviewing having to notice is far more appealing. 

As for your concern about main being outdated, well..worst case it updates with release PRs. It couldn't update main anyways due to branch protections

ckohen: copy copy

iCrawl (approved):

kyranet (approved):

PR #321, opened 2022-02-11
fix(APIInteraction): add modal submit interaction to union

almeidx:
- https://discord.com/channels/222078108977594368/941291502372552735/941658634973245531

### DIFF
--- a/deno/payloads/v8/interactions.ts
+++ b/deno/payloads/v8/interactions.ts
@@ -10,6 +10,7 @@ import type {
 	APIApplicationCommandInteraction,
 } from './_interactions/applicationCommands.ts';
 import type { APIApplicationCommandAutocompleteInteraction } from './_interactions/autocomplete.ts';
+import type { APIModalSubmitInteraction } from './_interactions/modalSubmit.ts';
 
 export * from './_interactions/base.ts';
 export * from './_interactions/messageComponents.ts';
@@ -25,7 +26,8 @@ export type APIInteraction =
 	| APIPingInteraction
 	| APIApplicationCommandInteraction
 	| APIMessageComponentInteraction
-	| APIApplicationCommandAutocompleteInteraction;
+	| APIApplicationCommandAutocompleteInteraction
+	| APIModalSubmitInteraction;
 
 /**
  * https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object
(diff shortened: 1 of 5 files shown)


vladfrangu (approved):
Nice catch

kyranet (approved):

tests/v9/interactions.test-d.ts
```diff
@@ -37,6 +38,10 @@ if (interaction.type === InteractionType.MessageComponent) {
 	}
 }
 
+if (interaction.type === InteractionType.ModalSubmit) {
+	expectType<APIModalSubmission | undefined>(interaction.data);
```
vladfrangu, line 42: Why would it be undefined... that doesn't seem right 👀 Can you make the tests be similar to the ones for ApplicationCommand? (as data shouldn't be undefined at all... and if it is, it's a bug we'll need to fix)

vladfrangu (changes requested):

tests/v9/interactions.test-d.ts
kyranet: ref: https://github.com/discord/discord-api-docs/pull/4486

almeidx: The docs are not clear about whether `data` is always present on modal submit interactions (~~I made that pull request~~ mostly just waiting on a staff response on the pr)

vladfrangu:
Might also need to check if it's similar to chat input commands with DM/Guild interactions 👀

deno/payloads/v8/interactions.ts
```diff
@@ -25,7 +26,8 @@ export type APIInteraction =
 	| APIPingInteraction
 	| APIApplicationCommandInteraction
 	| APIMessageComponentInteraction
-	| APIApplicationCommandAutocompleteInteraction;
+	| APIApplicationCommandAutocompleteInteraction
+	| APIModalSubmitInteraction;
```
vladfrangu, line 31: don't forget to update line 33 and 38 with the new DM/Guild ones :D

vladfrangu (changes requested):

vladfrangu (approved):

vladfrangu:
Looks great! Found the same missing guild/dm interactions for autocompletes that anyone can fix in a subsequent PR 👀

PR #329, opened 2022-02-14
chore(Changelog): show who created the pull request that was merged

vladfrangu:

Well, @suneettipirneni requested it, I wanted it too tbh, but I feel dirty for having achieved it like this...


iCrawl:
Does that even same save in the sense of not spamming the GH api? Or even worth it to do, listing all contributors, if one can see that natively in GH anyway.

vladfrangu:
Since it happens only on release PRs we're not spamming the API much.

For the showing concern, releases show contributors at the bottom when you tag them in the release body (you can see that for new contributors) and I think it would be nice to list everyone (and this is the cleanest solution, the alternative would be to use the generate changelog API and parse it out to add it to the release body)

iCrawl:
I'm just thinking that if we conform to some general approach, like using git-cliff, where we can't rely on another package, to have a solution that isn't tied to conventional-changelogs or so.

kyranet (approved):

PR #332, opened 2022-02-14
fix(APIApplicationCommandAutocompleteInteraction): make `options` field required

andre4ik3:
- Makes the `options` field required for autocomplete interactions
- Even though it is technically "optional" in the docs, in practice it will always be there when receiving an interaction
- The whole point of an autocomplete interaction is that a user is typing out an option. Therefore autocomplete interaction cannot be sent without options.

- [Interaction Object Data Structure](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object-interaction-data-structure)

### DIFF
--- a/tests/v9/interactions.test-d.ts
+++ b/tests/v9/interactions.test-d.ts
@@ -40,7 +40,10 @@ if (interaction.type === InteractionType.MessageComponent) {
 }
 
 if (interaction.type === InteractionType.ApplicationCommandAutocomplete) {
-	expectType<APIChatInputApplicationCommandInteractionData>(interaction.data);
+	expectType<
+		APIChatInputApplicationCommandInteractionData &
+			Required<Pick<APIChatInputApplicationCommandInteractionData, 'options'>>
+	>(interaction.data);
 }
 
 if (interaction.type === InteractionType.ModalSubmit) {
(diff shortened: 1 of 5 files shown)


andre4ik3:
Uh oh, I think there is issue with the test

andre4ik3:
Not sure if there is a better way to do this, I never used type testing before

tests/v9/interactions.test-d.ts
```diff
@@ -40,7 +40,10 @@ if (interaction.type === InteractionType.MessageComponent) {
 }
 
 if (interaction.type === InteractionType.ApplicationCommandAutocomplete) {
-	expectType<APIChatInputApplicationCommandInteractionData>(interaction.data);
+	expectType<
+		APIChatInputApplicationCommandInteractionData &
+			Required<Pick<APIChatInputApplicationCommandInteractionData, 'options'>>
```
vladfrangu, line 45: APIApplicationCommandAutocompleteInteraction['data']

vladfrangu (changes requested):

tests/v9/interactions.test-d.ts
andre4ik3: Awesome

vladfrangu (approved):

PR #343, opened 2022-02-23
Add `${bigint}` to Snowflake Type

IanMitchell:
Adds a second way of representing Snowflakes

### DIFF
--- a/deno/globals.ts
+++ b/deno/globals.ts
@@ -1,7 +1,7 @@
 /**
  * https://discord.com/developers/docs/reference#snowflakes
  */
-export type Snowflake = string;
+export type Snowflake = string | `${bigint}`;
 
 /**
  * https://discord.com/developers/docs/topics/permissions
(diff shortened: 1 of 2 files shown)


almeidx:
Is there a point in doing this? The type will just be narrowed to `string` regardless

[Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAygdgewO4DMA2BDA1tAvFAZ2ACcBLOAcygB8oADAEgG8AjUi84AXzoG4AoAPSCoogHoB+fvwAmEAMaZi0eQjhFCKAFyxEqTDgHDRJ0ycn8gA)

IanMitchell:
You are right! I was testing in a limbo state (rewrite) where it didn't simplify out this way - thought it was a problem here, but it was a problem in my lib. Thanks for pointing it out!

ImRodry:
The point in doing this is to make sure parameters that should take a snowflake only take a snowflake and nothing else. I think this should be added back, we really need this

PR #355, opened 2022-03-04
feat(Utils): add more typeguard functions to determine the interaction types

IanMitchell:
These functions are super useful for narrowing the type from the generic Interaction types into a specific interaction type

### DIFF
--- a/utils/v10.ts
+++ b/utils/v10.ts
@@ -18,6 +18,7 @@ import {
 	ApplicationCommandType,
 	ButtonStyle,
 	ComponentType,
+	InteractionType,
 } from '../payloads/v10/index';
 
 // Interactions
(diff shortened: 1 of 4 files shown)


utils/v10.ts
```diff
@@ -101,3 +107,53 @@ export function isLinkButton(component: APIButtonComponent): component is APIBut
...
 }
+
+// Message Components
+
+/**
+ * A type-guard check for button message component interactions
+ * @param interaction The message component interaction to check against
+ * @returns A boolean that indicates if the message component is a button
+ */
+export function isMessageComponentButtonInteraction(
```
vladfrangu, line 118: Can you also add a generic isMessageComponentInteraction? 🙏

vladfrangu (changes requested):

utils/v10.ts
IanMitchell: This is done

kyranet (approved):

vladfrangu (approved):

PR #357, opened 2022-03-05
chore: export APIButtonComponentBase<T> interface

IanMitchell:
This exports the APIButtonComponentBase interface. This is necessary for libraries or other projects that are defining abstracted button logic that is then extended into custom id or link buttons. 

You can see a [concrete usage of this type here](https://github.com/IanMitchell/interaction-kit/pull/108/commits/1d8437b260d2be6412185b6dacbaf7c3835d7feb).


kyranet (approved):

vladfrangu (approved):
tfw we were exporting basecomponent but not basebutton.

PR #366, opened 2022-03-11
fix(APIGuildIntegrationType): correct name of type

Jiralite:
The type of an integration was exported as `APIGuildInteractionType`, but this is misnamed. It should be `APIGuildIntegrationType`. It's not an interaction.


vladfrangu:
Actually screaming

vladfrangu (approved):

PR #402, opened 2022-04-16
docs: fix links from DDevs portal

vitorlops:
This PR changes a bunch of links that were wrong.

### DIFF
--- a/deno/payloads/v6/channel.ts
+++ b/deno/payloads/v6/channel.ts
@@ -131,7 +131,7 @@ export interface APIMessageApplication {
 }
 
 /**
- * https://discord.com/developers/docs/resources/channel#message-object-message-reference-structure
+ * https://discord.com/developers/docs/resources/channel#message-reference-object-message-reference-structure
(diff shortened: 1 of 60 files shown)


deno/payloads/v6/channel.ts
```diff
@@ -131,7 +131,7 @@ export interface APIMessageApplication {
 }
 
 /**
- * https://discord.com/developers/docs/resources/channel#message-object-message-reference-structure
+ * https://discord.com/developers/docs/resources/channel#message-reference-object-message-reference-structure
```
vladfrangu, line 134: v6 and v8 don't get updates

vladfrangu (changes requested):

deno/payloads/v6/channel.ts
vitorlops: I thought it would be good, as they are links anyway. Be aware that it won't work for people using these vesions.
But if you disagree I can revert it.

vladfrangu (approved):

PR #404, opened 2022-04-19
chore(ActivityPlatform): update platform enum

advaith1:


deno/payloads/v10/gateway.ts
```diff
@@ -152,7 +152,10 @@ export interface GatewayActivity {
 	 */
 	timestamps?: GatewayActivityTimestamps;
 	sync_id?: string;
-	platform?: ActivityPlatform;
+	/**
+	 * {@link ActivityPlatform}
```
vladfrangu, line 156: 1. Does it work
2. You should use something more descriptive (`You can use {@link ActivityPlatform} as a stepping stone, but this might be inaccurate`)

vladfrangu (changes requested):

almeidx:
Shouldn't the `id`, `sync_id`, `platform`, and `session_id` be removed/have a `@unstable` comment? They aren't documented
https://discord.com/developers/docs/topics/gateway#activity-object

I mentioned `platform` because it's not just the values (enum) that aren't documented, the property itself is not documented at all

advaith1:
I use `id`, `platform`, and `sync_id`.

deno/payloads/v10/gateway.ts
advaith1: > but this might be inaccurate

tbh this is true for all enums since discord adds new values all the time ¯\\\_(ツ)_/¯

vladfrangu (approved):

PR #417, opened 2022-04-30
fix(OAuth2Scopes): update ApplicationsCommandsUpdate

john-batch:
The OAuth2 scope that allows an integration to update it's command permissions recently changed from `applications.commands.update` to `applications.commands.permissions.update` : https://discord.com/developers/docs/topics/oauth2#shared-resources-oauth2-scopes


advaith1:
no, those two scopes are totally separate

john-batch:
Oh my bad, the link I posted does mention that as well, I should have read it more carefully. Sorry for the inconvience.

PR #495, opened 2022-06-17
feat(UserFlags): add `Quarantined` flag

Mateo-tem:
Adds `Quarantined (1 << 44)` flag to UserFlags.

### DIFF
--- a/deno/payloads/v10/user.ts
+++ b/deno/payloads/v10/user.ts
@@ -145,6 +145,10 @@ export enum UserFlags {
 	 * User has been identified as spammer
 	 */
 	Spammer = 1 << 20,
+	/**
+	 * User's account has been quarantined based on recent activity
+	 */
+	Quarantined = 1 << 44
 }
 
 /**
--- a/deno/payloads/v10/user.ts
+++ b/deno/payloads/v10/user.ts
@@ -148,7 +148,7 @@ export enum UserFlags {
 	/**
 	 * User's account has been quarantined based on recent activity
 	 */
-	Quarantined = 1 << 44
+	Quarantined = Math.pow(2, 44),
 }
 
 /**
--- a/deno/payloads/v10/user.ts
+++ b/deno/payloads/v10/user.ts
@@ -146,7 +146,7 @@ export enum UserFlags {
 	 */
 	Spammer = 1 << 20,
 	/**
-	 * User's account has been quarantined based on recent activity
+	 * @unstable User's account has been quarantined based on recent activity
 	 */
 	Quarantined = Math.pow(2, 44),
 }
--- a/deno/payloads/v10/user.ts
+++ b/deno/payloads/v10/user.ts
@@ -146,7 +146,9 @@ export enum UserFlags {
 	 */
 	Spammer = 1 << 20,
 	/**
-	 * @unstable User's account has been quarantined based on recent activity
+	 * User's account has been quarantined based on recent activity
+	 *
+	 * @unstable This user flag is currently not documented by Discord but has known values which we will try to keep up to date.
 	 */
 	Quarantined = Math.pow(2, 44),
 }
--- a/deno/payloads/v10/user.ts
+++ b/deno/payloads/v10/user.ts
@@ -148,7 +148,7 @@ export enum UserFlags {
 	/**
 	 * User's account has been quarantined based on recent activity
 	 *
-	 * @unstable This user flag is currently not documented by Discord but has known values which we will try to keep up to date.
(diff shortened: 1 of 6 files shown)


deno/payloads/v10/user.ts
```diff
@@ -145,6 +145,10 @@ export enum UserFlags {
 	 * User has been identified as spammer
 	 */
 	Spammer = 1 << 20,
+	/**
+	 * User's account has been quarantined based on recent activity
+	 */
+	Quarantined = 1 << 44
```
kyranet, line 151: This is invalid, `1 << 44` yields the same value as `1 << 12`.

kyranet (dismissed):

deno/payloads/v10/user.ts
vladfrangu: Is it bigints time? 🥲

almeidx: The client uses `Math.pow(2, 44)`

vladfrangu: That's webpack mostly. We should still aim to use bitshifting

vvito7: It should also use `@unstable`, since it wasn't documented by Discord

Mateo-tem: > It should also use `@unstable`, since it wasn't documented by Discord

I think it wouldn't be documented in Discord API Docs, in the same way as the `Spammer` user flag.

I just have this reference: https://support.discord.com/hc/en-us/articles/6461420677527

almeidx: The Spammer userflag should have an `@unstable` comment as well

kyranet: > The client uses `Math.pow(2, 44)`

Well, that does work:
image

However, it's a bit useless as integers, because you can't bitshift bits higher than 32 bits.

vladfrangu: Let's use the Math.pow for now (or directly the number for the bit) and when Discord moves to bigints, so will we

Mateo-tem: Done.

almeidx:
Can you add `@unstable` to it, since it's not documented

deno/payloads/v10/user.ts
```diff
@@ -145,6 +145,10 @@ export enum UserFlags {
 	 * User has been identified as spammer
 	 */
 	Spammer = 1 << 20,
+	/**
+	 * @unstable User's account has been quarantined based on recent activity
+	 */
```
vvito7, line 150:
```suggestion
	/**
	 * User's account has been quarantined based on recent activity
	 *
	 * @unstable This User flag is currently not documented by Discord but has known values which we will try to keep up to date.
	 */
```
Probably should also apply to Spammer, as said in https://github.com/discordjs/discord-api-types/pull/495#discussion_r900419075.

deno/payloads/v8/user.ts
```diff
@@ -148,7 +148,9 @@ export enum UserFlags {
 	 */
 	Spammer = 1 << 20,
 	/**
-	 * @unstable User's account has been quarantined based on recent activity
+	 * User's account has been quarantined based on recent activity
+	 *
+	 * @unstable This user flag is currently not documented by Discord but has known values which we will try to keep up to date.
```
vladfrangu, line 153: Can you add this to the spammer flag too

vladfrangu (changes requested):

deno/payloads/v8/user.ts
Mateo-tem: I was about to do that, but I don't know if it would be better if I open another PR with that.

vladfrangu: That's fine with me!

Mateo-tem: Done! #500

deno/payloads/v10/user.ts
```diff
@@ -145,6 +145,12 @@ export enum UserFlags {
 	 * User has been identified as spammer
 	 */
 	Spammer = 1 << 20,
+	/**
+	 * User's account has been quarantined based on recent activity
+	 *
+	 * @unstable This user flag is currently not documented by Discord but has known values which we will try to keep up to date.
```
vladfrangu, line 151:
```suggestion
	 * @unstable This user flag is currently not documented by Discord but has a known value which we will try to keep up to date.
```

vladfrangu (changes requested):

vladfrangu (approved):

PR #453, opened 2022-05-27
feat(routes): add guild member banner route

suneettipirneni:
**Please describe the changes this PR makes and why it should be merged:**

**If applicable, please reference Discord API Docs PRs or commits that influenced this PR:**

- https://github.com/discord/discord-api-docs/pull/4955

PR #502, opened 2022-06-20
feat(REST): add `CDNRoutes`

vvito7:
Adds the constant `CDNRoutes`, that contains each CDN entpoint from [DDevs Portal](https://discord.com/developers/docs/reference#image-formatting-cdn-endpoints). Made it because of https://github.com/discordjs/discord-api-types/pull/453#issuecomment-1139164200. With this, many other things are being added: `ImageFormats`, `CDNQuery` and `*Format` types.

### DIFF
--- a/deno/rest/v10/mod.ts
+++ b/deno/rest/v10/mod.ts
@@ -844,6 +844,180 @@ export const Routes = {
 	},
 };
 
+export const CDNRoutes = {
+	/**
+	 * Route for:
+	 * - GET `/emojis/{emoji.id}.{png|jpeg|webp|gif}`
+	 *
+	 * As this route supports GIFs, the hash will begin with `a_` if it is available in GIF format
+	 */
+	customEmoji(emojiId: Snowflake, format: 'png' | 'jpeg' | 'webp' | 'gif') {
+		return `/emojis/${emojiId}.${format}` as const;
+	},
+
+	/**
+	 * Route for:
+	 * - GET `/guilds/{guild.id}/icons/{guild.id}.{png|jpeg|webp|gif}`
+	 *
+	 * As this route supports GIFs, the hash will begin with `a_` if it is available in GIF format
+	 */
+	guildIcon(guildId: Snowflake, guildIcon: string, format: 'png' | 'jpeg' | 'webp' | 'gif') {
+		return `icons/${guildId}/${guildIcon}.${format}` as const;
+	},
+
+	/**
+	 * Route for:
+	 * - GET `/splashes/{guild.id}/{guild.splash}.{png|jpeg|webp}`
+	 */
+	guildSplash(guildId: Snowflake, guildSplash: string, format: 'png' | 'jpeg' | 'webp') {
+		return `/splashes/${guildId}/${guildSplash}.${format}` as const;
+	},
+
+	/**
+	 * Route for:
+	 * - GET `/discovery-splashes/{guild.id}/{guild.discovery_splash}.{png|jpeg|webp}`
+	 */
+	guildDiscoverySplash(guildId: Snowflake, guildDiscoverySplash: string, format: 'png' | 'jpeg' | 'webp') {
+		return `/discovery-splashes/${guildId}/${guildDiscoverySplash}.${format}` as const;
+	},
+
+	/**
+	 * Route for:
+	 * - GET `/banners/{guild.id}/{guild.banner}.{png|jpeg|webp|gif}`
+	 *
+	 * As this route supports GIFs, the hash will begin with `a_` if it is available in GIF format
+	 */
+	guildBanner(guildId: Snowflake, guildBanner: string, format: 'png' | 'jpeg' | 'webp' | 'gif') {
+		return `/banners/${guildId}/${guildBanner}.${format}` as const;
+	},
+
+	/**
+	 * Route for:
+	 * - GET `/banners/{user.id}/{user.banner}.{png|jpeg|webp|gif}`
+	 *
+	 * As this route supports GIFs, the hash will begin with `a_` if it is available in GIF format
+	 */
+	userBanner(userId: Snowflake, userBanner: string, format: 'png' | 'jpeg' | 'webp' | 'gif') {
+		return `/banners/${userId}/${userBanner}.${format}` as const;
+	},
+
+	/**
+	 * Route for:
+	 * - GET `/embed/avatars/{user.discriminator % 5}.png`
+	 *
+	 * The `userDiscriminator` parameter should be the user discriminator modulo 5 (e.g. 1337 % 5 = 2)
+	 */
+	defaultUserAvatar(userDiscriminator: Snowflake) {
+		return `/embed/avatars/${userDiscriminator}.png` as const;
+	},
+
+	/**
+	 * Route for:
+	 * - GET `/avatars/{user.id}/{user.avatar}.{png|jpeg|webp|gif}`
+	 *
+	 * As this route supports GIFs, the hash will begin with `a_` if it is available in GIF format
+	 */
+	userAvatar(userId: Snowflake, userAvatar: string, format: 'png' | 'jpeg' | 'webp' | 'gif') {
+		return `/avatars/${userId}/${userAvatar}.${format}` as const;
+	},
+
+	/**
+	 * Route for:
+	 * - GET `/guilds/{guild.id}/users/{user.id}/{guild_member.avatar}.{png|jpeg|webp|gif}`
+	 *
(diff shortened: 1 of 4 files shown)


deno/rest/v10/mod.ts
```diff
@@ -844,6 +844,180 @@ export const Routes = {
...
+		return `/banners/${userId}/${userBanner}.${format}` as const;
+	},
+
+	/**
+	 * Route for:
+	 * - GET `/embed/avatars/{user.discriminator % 5}.png`
+	 *
+	 * The `userDiscriminator` parameter should be the user discriminator modulo 5 (e.g. 1337 % 5 = 2)
+	 */
+	defaultUserAvatar(userDiscriminator: Snowflake) {
```
vladfrangu, line 910: Its not a snowflake

vladfrangu (changes requested):

deno/rest/v10/mod.ts
```diff
@@ -844,6 +844,180 @@ export const Routes = {
...
 };
 
+export const CDNRoutes = {
+	/**
+	 * Route for:
+	 * - GET `/emojis/{emoji.id}.{png|jpeg|webp|gif}`
+	 *
+	 * As this route supports GIFs, the hash will begin with `a_` if it is available in GIF format
+	 */
+	customEmoji(emojiId: Snowflake, format: 'png' | 'jpeg' | 'webp' | 'gif') {
```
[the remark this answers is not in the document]
vvito7: The function name, right?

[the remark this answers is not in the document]
vladfrangu: Yea

vladfrangu:
> this PR doesn't add the last added CDN endpoint

It should, as the PR you linked adds it as a route, not a cdn route

deno/rest/v10/mod.ts
```diff
@@ -844,6 +844,182 @@ export const Routes = {
...
 };
 
+export const CDNRoutes = {
+	/**
+	 * Route for:
+	 * - GET `/emojis/{emoji.id}.{png|jpeg|webp|gif}`
+	 *
+	 * As this route supports GIFs, the hash will begin with `a_` if it is available in GIF format
+	 */
+	emoji(emojiId: Snowflake, format: 'png' | 'jpeg' | 'webp' | 'gif') {
```
vladfrangu, line 854: Can you export a type for the formats? Or better yet make them an enum

deno/rest/v10/mod.ts
```diff
@@ -844,6 +844,182 @@ export const Routes = {
...
+
+	/**
+	 * Route for:
+	 * - GET `/embed/avatars/{user.discriminator % 5}.png`
+	 *
+	 * The `userDiscriminator` parameter should be the user discriminator modulo 5 (e.g. 1337 % 5 = 2)
+	 *
+	 * The numbers range from only 0 to 5
+	 */
+	defaultUserAvatar(userDiscriminator: number) {
```
vladfrangu, line 912:
```suggestion
	defaultUserAvatar(userDiscriminator: 0 | 1 | 2 | 3 | 4 | 5) {
```

vladfrangu (changes requested):

deno/rest/v10/mod.ts
vvito7: I tried this for it limit the formats, but i get this error:
image
image

So there is no limitation, but I added a more descripitive JSDoc:
Example: `This route supports the extensions: PNG, JPEG, WebP, GIF`

almeidx: `Exclude<ImageFormat, ImageFormat.Lottie>`

ImRodry (commented):
Don't you wanna add a size parameter in every endpoint?

vladfrangu:
Outside of defining them as interfaces, just like how we do for other rest endpoints, no. So, if @vvito7 wants to do that then they can, otherwise this is fineee

didinele:
Should we maybe export & use types like `EmojiFormat = Exclude<ImageFormat, ImageFormat.Lottie>` instead of typing it out everywhere? It's more maintainable for both us and the end-user (i.e. if they write a function that takes a format it then passes it into our methods)

vvito7:
> Outside of defining them as interfaces, just like how we do for other rest endpoints, no. So, if @vvito7 wants to do that then they can, otherwise this is fineee

I could of course, but would it be something like this? I got a little confused
```ts
export interface RESTGetCDNRouteQuery {
	/**
	 * The returned image can have the size changed by using this query parameter
	 */
	size?: AllowedImageSizes;
}

export type AllowedImageSizes = 16 | 32 | 56 | 64 | 96 | 128 | 256 | 300 | 512 | 600 | 1024 | 2048 | 4096;
```

> Should we maybe export & use types like `EmojiFormat = Exclude<ImageFormat, ImageFormat.Lottie>` instead of typing it out everywhere? It's more maintainable for both us and the end-user (i.e. if they write a function that takes a format it then passes it into our methods)

I like the idea 👍

vladfrangu:
> > Outside of defining them as interfaces, just like how we do for other rest endpoints, no. So, if @vvito7 wants to do that then they can, otherwise this is fineee
> 
> I could of course, but would it be something like this? I got a little confused
> 
> ```ts
> export interface RESTGetCDNRouteQuery {
> 	/**
> 	 * The returned image can have the size changed by using this query parameter
> 	 */
> 	size?: AllowedImageSizes;
> }
> 
> export type AllowedImageSizes = 16 | 32 | 56 | 64 | 96 | 128 | 256 | 300 | 512 | 600 | 1024 | 2048 | 4096;
> ```

Yes, but maybe we should just type size as a number? Also this can use a diff interface name (CDNQuery?) if you want

vvito7:
> maybe we should just type size as a number?

Yes, sure. But there is a bunch of numbers that don't work


> Also this can use a diff interface name (CDNQuery?) if you want

Sure👍

vladfrangu:
> I could just add the allowed sizes in the docs

Well its just powers of 2 so... that should be easy enough

ImRodry:
Except it isn’t

vladfrangu:
Source?

ImRodry:
Discord.js, discord docs, whatever floats your boat

vladfrangu:
Well call me nuts because Discord says it themselves: https://discord.com/developers/docs/reference#image-formatting-image-base-url

> The returned size can be changed by appending a querystring of `?size=desired_size` to the URL. Image size can be any power of two between 16 and 4096.

didinele:
That's basically it, just make each one a `type` rather than putting it under an interface, e.g.
`type Emoji = Exclude<ImageFormat, ImageFormat.Lottie>;`

> Also there is so many duplicated stuff
that's alright

didinele (approved):

kyranet (approved):

deno/rest/v10/mod.ts
```diff
@@ -844,6 +844,258 @@ export const Routes = {
...
+
+	/**
+	 * Route for:
+	 * - GET `/embed/avatars/{user.discriminator % 5}.png`
+	 *
+	 * The `userDiscriminator` parameter should be the user discriminator modulo 5 (e.g. 1337 % 5 = 2)
+	 *
+	 * This route supports the extension: PNG
+	 */
+	defaultUserAvatar(userDiscriminator: 0 | 1 | 2 | 3 | 4 | 5) {
```
vladfrangu, line 924: make these a type too

deno/rest/v10/mod.ts
```diff
@@ -844,6 +844,258 @@ export const Routes = {
...
+	},
+
+	/**
+	 * Route for:
+	 * - GET `/app-assets/710982414301790216/store/{sticker_pack.banner.asset_id}.{png|jpeg|webp}`
+	 *
+	 * This route supports the extensions: PNG, JPEG, WebP
+	 */
+	stickerPackBanner(stickerPackBannerAssetId: Snowflake, format: StickerPackBannerFormat) {
+		return `app-assets/710982414301790216/store/${stickerPackBannerAssetId}.${format}` as const;
```
vladfrangu, line 1004:
```suggestion
		return `/app-assets/710982414301790216/store/${stickerPackBannerAssetId}.${format}` as const;
```

vladfrangu (changes requested):

deno/rest/v10/mod.ts
almeidx: Could the app id also be a constant somewhere (710982414301790216)? Something like `StickerPackApplicationId` maybe

vvito7: This const had to be placed on the top of `CDNRoutes`

vladfrangu (approved):

PR #539, opened 2022-07-24
fix(OverwriteType): change from number to string

thehairy:

The API returns the overwrite type as strings, not numbers, already since v8.
https://discord.com/developers/docs/change-log#api-and-gateway-v8


vvito7:
it uses string in the audit log object but number in the channel object

PR #556, opened 2022-08-06
docs: add notes about `MESSAGE_CONTENT` intent in affected data fields

Mateo-tem:

Adds notes about the `MESSAGE_CONTENT` intent, which will be [privileged for verified apps after August 31, 2022](https://support-dev.discord.com/hc/en-us/articles/4404772028055).
There are affected fields in `APIMessage`, which are `content`, `embeds`, `attachments`, and `components`.

### DIFF
--- a/deno/payloads/v10/channel.ts
+++ b/deno/payloads/v10/channel.ts
@@ -345,8 +345,7 @@ export interface APIMessage {
 	/**
 	 * Contents of the message
 	 *
-	 * `MESSAGE_CONTENT` (`1 << 15`) will be a privileged intent for verified apps,
-	 * this intent is required to receive a non-empty value from this field
+	 * The `MESSAGE_CONTENT` (`1 << 15`) privileged gateway intent is required for verified applications, to receive a non-empty value from this field
 	 *
 	 * See https://support-dev.discord.com/hc/en-us/articles/4404772028055
 	 */
@@ -400,8 +399,7 @@ export interface APIMessage {
 	 *
 	 * See https://discord.com/developers/docs/resources/channel#attachment-object
 	 *
-	 * `MESSAGE_CONTENT` (`1 << 15`) will be a privileged intent for verified apps,
-	 * this intent is required to receive a non-empty value from this field
+	 * The `MESSAGE_CONTENT` (`1 << 15`) privileged gateway intent is required for verified applications, to receive a non-empty value from this field
 	 *
 	 * See https://support-dev.discord.com/hc/en-us/articles/4404772028055
 	 */
@@ -411,8 +409,7 @@ export interface APIMessage {
 	 *
 	 * See https://discord.com/developers/docs/resources/channel#embed-object
 	 *
-	 * `MESSAGE_CONTENT` (`1 << 15`) will be a privileged intent for verified apps,
-	 * this intent is required to receive a non-empty value from this field
+	 * The `MESSAGE_CONTENT` (`1 << 15`) privileged gateway intent is required for verified applications, to receive a non-empty value from this field
 	 *
 	 * See https://support-dev.discord.com/hc/en-us/articles/4404772028055
 	 */
@@ -499,8 +496,7 @@ export interface APIMessage {
 	/**
 	 * Sent if the message contains components like buttons, action rows, or other interactive components
 	 *
-	 * `MESSAGE_CONTENT` (`1 << 15`) will be a privileged intent for verified apps,
-	 * this intent is required to receive a non-empty value from this field
+	 * The `MESSAGE_CONTENT` (`1 << 15`) privileged gateway intent is required for verified applications, to receive a non-empty value from this field
 	 *
 	 * See https://support-dev.discord.com/hc/en-us/articles/4404772028055
 	 */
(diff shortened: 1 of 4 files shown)


Jiralite (commented):
For API version 10, this is already in effect, so is there a need to specify the date? I do not think so?

SpaceEEC (approved):

Jiralite (changes requested):
>	 * `MESSAGE_CONTENT` (`1 << 15`) will be a privileged intent for verified apps,
>	 * this intent is required to receive a non-empty value from this field

I honestly think this doesn't flow smoothly and still doesn't quite make sense for version 10. For version 10, this is already in effect. This isn't a _will be_, it _already is_.

For version 10, I would recommend something like this:

> The `MESSAGE_CONTENT` (`1 << 15`) privileged gateway intent is required for verified applications to receive a non-empty value from this field

For versions before this, I recommend something like this:

> The `MESSAGE_CONTENT` (`1 << 15`) privileged gateway intent will become required [date here] for verified applications to receive a non-empty value from this field

kyranet (approved):

deno/payloads/v10/channel.ts
```diff
@@ -344,6 +344,10 @@ export interface APIMessage {
 	author: APIUser;
 	/**
 	 * Contents of the message
+	 *
+	 * The `MESSAGE_CONTENT` (`1 << 15`) privileged gateway intent is required for verified applications, to receive a non-empty value from this field
```
vladfrangu, line 348: That comma feels wrong...

deno/payloads/v9/channel.ts
```diff
@@ -344,6 +344,10 @@ export interface APIMessage {
 	author: APIUser;
 	/**
 	 * Contents of the message
+	 *
+	 * The `MESSAGE_CONTENT` (`1 << 15`) privileged gateway intent will become required after August 31, 2022 for verified applications, to receive a non-empty value from this field
```
vladfrangu, line 348: Isn't this applicable only to API v10?

vladfrangu (changes requested):

deno/payloads/v9/channel.ts
advaith1: currently its only enforced on v10, after the deadline it will be enforced on all API versions

however for REST it's pointless to put the bit value, since it only depends on the dev portal toggle

Mateo-tem: (From what I understood) API v10 needs to put the bit value when connecting to the gateway, and have the toggle from Discord Developers Portal enabled, but in API v9, **only the toggle** needs to be enabled. Is it correct?

advaith1: oh this is in payloads; yeah i think that's right

Mateo-tem: Oh, right. Maybe I can change the descriptions with this information for both API versions, so as to not generalize.
In API v10 can be like:

> The **`MESSAGE_CONTENT`** privileged gateway intent is required for verified applications to receive a non-empty value from this field. In Discord Developers Portal, you need to enable the toggle of this intent of your application in **Bot > Privileged Gateway Intents**. You also need to specify the intent bit value  (**`1 << 15`**) when connecting to the gateway.

and for API v9 can be like:

> The **`MESSAGE_CONTENT`** privileged gateway intent will become required after **August 31, 2022** for verified applications to receive a non-empty value from this field. In Discord Developers Portal, you need to enable the toggle of this intent of your application in **Bot > Privileged Gateway Intents**.

vladfrangu: > > The **`MESSAGE_CONTENT`** privileged gateway intent is required for verified applications to receive a non-empty value from this field. In Discord Developers Portal, you need to enable the toggle of this intent of your application in **Bot > Privileged Gateway Intents**. You also need to specify the intent bit value (**`1 << 15`**) when connecting to the gateway.

I'd correct this to `You also need to specify the intent bit value...if you are connecting to the gateway`

deno/payloads/v10/channel.ts
```diff
@@ -344,6 +344,13 @@ export interface APIMessage {
 	author: APIUser;
 	/**
 	 * Contents of the message
+	 *
+	 * The **`MESSAGE_CONTENT`** privileged gateway intent is required for verified applications to receive a non-empty value from this field
+	 *
+	 * In Discord Developers Portal, you need to enable the toggle of this intent of your application in **Bot > Privileged Gateway Intents**.
+	 * You also need to specify the intent bit value  (**`1 << 15`**) if you are connecting to the gateway
+	 *
+	 * See https://support-dev.discord.com/hc/en-us/articles/4404772028055
```
Jiralite, line 353: Bolding in-line code has no effect. Also removed a double space and added "In **the** Discord Developers Portal...":

```suggestion
	 *
	 * The `MESSAGE_CONTENT` privileged gateway intent is required for verified applications to receive a non-empty value from this field
	 *
	 * In the Discord Developers Portal, you need to enable the toggle of this intent of your application in **Bot > Privileged Gateway Intents**.
	 * You also need to specify the intent bit value (`1 << 15`) if you are connecting to the gateway
	 *
	 * See https://support-dev.discord.com/hc/en-us/articles/4404772028055
```

PR #557, opened 2022-08-07
fix: Oauth2Routes

grarich:

### DIFF
--- a/deno/rest/v10/mod.ts
+++ b/deno/rest/v10/mod.ts
@@ -1101,6 +1101,7 @@ export interface CDNQuery {
 }
 
 export const RouteBases = {
+	apiBase: 'https://discord.com/api',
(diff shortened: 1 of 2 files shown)


vladfrangu:
Are you sure this is correct? Pretty sure even this route is versioned

grarich:
Thanks for the reply.
Please see here.
https://discord.com/developers/docs/topics/oauth2

vladfrangu:
I mean I know they don't list the versioning but the domain definitely supports versioning, and especially switching between v6 and v7+ shows different, usable errors 👀 
Probably needs clarification from discords side for this

grarich:
I have tried accessing the site by specifying a version such as v6, but I am redirected to a page with no version control.
I don't know because I don't have much access to the Oauth2 API, but was it version controlled before?

PR #560, opened 2022-08-10
feat(APIApplicationCommandIntegerOption): add `min_length` and `max_length`

dager-mohamed:


vladfrangu:
W-what?

dager-mohamed:
> W-what?

if you see in https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-structure you will see `min-length` and `max-length` for command options

vladfrangu:
Yes...only for string options, which is not where you added it...

dager-mohamed:
> Yes...only for string options, which is not where you added it...

bro it is integer!

dager-mohamed:
> Yes...only for string options, which is not where you added it...

image

dager-mohamed:
> Yes...only for string options, which is not where you added it...

lol I added as a boolean in PR not an integer lol

vladfrangu:
In fact, for integer/number options you use min/max_value (which are already present). min/max_length is only for string options (where it makes sense) 🙂 

The screenshot you sent refers to the data type you send to discord when specifying the option, *not* the option type! In this case, Discord is telling you that if you're specifying min/max_length the value needs to be an integer

dager-mohamed:
> Yes...only for string options, which is not where you added it...

anyway I have been edited the pull request

dager-mohamed:
> In fact, for integer/number options you use min/max_value (which are already present). min/max_length is only for string options (where it makes sense) slightly_smiling_face
> 
> The screenshot you sent refers to the data type you send to discord when specifying the option, _not_ the option type! In this case, Discord is telling you that if you're specifying min/max_length the value needs to be an integer

lol ok thx bro

PR #563, opened 2022-08-12
feat(APIGuildIntegration): add `scopes`

vvito7:

- https://github.com/discord/discord-api-docs/pull/5311


ooliver1 (changes requested):
> Unsure about optional/nullable since I don't have access to a youtube/twitch integration to test with, just assuming it's optional.

So you should make scopes nullable either via `?` or `null` depending on what kind of optional it is.

kyranet (approved):

PR #588, opened 2022-09-15
pascal guild integration types

deliever42:
i changed the guild integration types to pascal


vladfrangu (changes requested):

vladfrangu:
You have CI errors! Also, as said before, you should keep the old type too and add an `@deprecated` tag

PR #624, opened 2022-10-27
fix(UserFlags): hardcode the value of `Quarantined`

almeidx:
The tsc is not able to compute the value of `Math.pow(2, 44)` at compile-time, leaving the `Quarantined` enum member without an initializer in the .d.ts file, which prevents thing such as:
image

### DIFF
--- a/payloads/v10/user.ts
+++ b/payloads/v10/user.ts
@@ -151,8 +151,12 @@ export enum UserFlags {
 	 * User's account has been quarantined based on recent activity
 	 *
 	 * @unstable This user flag is currently not documented by Discord but has a known value which we will try to keep up to date.
+	 *
+	 * @privateRemarks
+	 *
+	 * This value would be 1 << 44, but bit shifting above 1 << 30 requires bigints
 	 */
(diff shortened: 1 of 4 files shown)


payloads/v10/user.ts
```diff
@@ -151,8 +151,12 @@ export enum UserFlags {
 	 * User's account has been quarantined based on recent activity
 	 *
 	 * @unstable This user flag is currently not documented by Discord but has a known value which we will try to keep up to date.
+	 *
+	 * @privateRemarks
+	 *
+	 * This value would be 1 << 44, but bit shifting above 1 << 30 requires bigints
```
kyranet, line 157: Can't we keep this on the same line just like with `@unstable`?

almeidx: According to tsdoc, it's a block tag (which means: no)
https://tsdoc.org/pages/tags/privateremarks/

kyranet (approved):

vladfrangu (approved):

advaith1:
Is the ~ at the end of the line intentional?

PR #666, opened 2022-12-10
fix(APIChannel): correctly type `name` based on channel type

cyyynthia:
Types incorrectly assume PartialChannel objects will always have a name.
In some circumstances, PartialChannels may be DM channels, which do have a null name.

Also fixes the comment about minimum name length.

This change aligns with the already documented [Channel data structure](https://discord.com/developers/docs/resources/channel#channel-object-channel-structure)

### DIFF
--- a/deno/payloads/v10/channel.ts
+++ b/deno/payloads/v10/channel.ts
@@ -73,7 +73,7 @@ export interface APIGuildChannel<T extends ChannelType> extends Omit<APIChannelB
 	/**
 	 * The name of the channel (1-100 characters)
 	 */
-	name?: string;
+	name: string;
 	/**
 	 * The id of the guild (may be missing for some channel objects received over gateway guild dispatches)
 	 */
@@ -169,10 +169,14 @@ export interface APIDMChannel extends Omit<APIDMChannelBase<ChannelType.DM>, 'na
 	/**
 	 * The name of the channel (always null for DM channels)
 	 */
-	name?: null;
+	name: null;
 }
 
-export interface APIGroupDMChannel extends APIDMChannelBase<ChannelType.GroupDM> {
+export interface APIGroupDMChannel extends Omit<APIDMChannelBase<ChannelType.GroupDM>, 'name'> {
+	/**
+	 * The name of the channel (1-100 characters)
+	 */
+	name: string | null;
 	/**
 	 * Application id of the group DM creator if it is bot-created
 	 */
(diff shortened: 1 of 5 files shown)


vladfrangu:
Can you apply it to the other channel types to make name non-nullable 🙏

cyyynthia:
Done, I've made the Guild/DM channel types more specific regarding whether their `name` is nullable or not. DM channels have it always `null`, GroupDM inherit the generic type (`string | null`), guild channels have it always `string`

vladfrangu:
Just so we're safe there's no regressions, can you add some tests too please? 🥺

cyyynthia:
> can you add some tests too please?

Done, I've only tested the `name` property though

cyyynthia:
I'm quite unsure as to why the CI build is failing telling me I didn't build the Deno types, `npm run build:deno` does 0 changes and the job's logs seem weird considering the changes I've made.. bad check script, or an oopsie on my end? 🤔

deno/payloads/v10/channel.ts
```diff
@@ -69,7 +69,11 @@ export interface APITextBasedChannel<T extends ChannelType> extends APIChannelBa
 	rate_limit_per_user?: number;
 }
 
-export interface APIGuildChannel<T extends ChannelType> extends APIChannelBase<T> {
+export interface APIGuildChannel<T extends ChannelType> extends Omit<APIChannelBase<T>, 'name'> {
+	/**
+	 * The name of the channel (1-100 characters)
+	 */
+	name?: string;
```
vladfrangu, line 76: Guild channels will always have a name

deno/payloads/v10/channel.ts
```diff
@@ -161,9 +165,14 @@ export interface APIDMChannelBase<T extends ChannelType> extends Omit<APITextBas
 	recipients?: APIUser[];
 }
 
-export type APIDMChannel = APIDMChannelBase<ChannelType.DM>;
+export interface APIDMChannel extends Omit<APIDMChannelBase<ChannelType.DM>, 'name'> {
+	/**
+	 * The name of the channel (always null for DM channels)
```
vladfrangu, line 170: Is it always returned as null or can it be undefined too? If it can't be undefined, drop the `?`

vladfrangu (changes requested):

cyyynthia:
oof, GitHub's re-request thing dropped the other reviewers 💀

vladfrangu (approved):

PR #669, opened 2022-12-12
fix(APIChannel): correctly type present properties based on channel type

cyyynthia:
Sort-of follow up of #666 after finding out some additional problems with the typings of the `APIChannel`-related objects

 - `default_thread_rate_limit_per_user` can be set on regular channels, despite the Discord client not showing this possibility, and newly created threads do abide by this configuration.
 - `last_pin_timestamp` was not present on DM channels, despite them having this property.

Tests included

N/A

### DIFF
--- a/deno/payloads/v10/channel.ts
+++ b/deno/payloads/v10/channel.ts
@@ -58,6 +58,11 @@ export interface APITextBasedChannel<T extends ChannelType> extends APIChannelBa
 	 */
 	last_message_id?: Snowflake | null;
 	/**
+	 * When the last pinned message was pinned.
+	 * This may be `null` in events such as `GUILD_CREATE` when a message is not pinned
+	 */
+	last_pin_timestamp?: string | null;
+	/**
 	 * Amount of seconds a user has to wait before sending another message (0-21600);
 	 * bots, as well as users with the permission `MANAGE_MESSAGES` or `MANAGE_CHANNELS`, are unaffected
 	 *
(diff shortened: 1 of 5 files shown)


deno/payloads/v10/channel.ts
```diff
@@ -196,7 +201,7 @@ export interface APIGroupDMChannel extends Omit<APIDMChannelBase<ChannelType.Gro
 }
 
 export interface APIThreadChannel
-	extends APIGuildChannel<ChannelType.PublicThread | ChannelType.PrivateThread | ChannelType.AnnouncementThread> {
+	extends APITextBasedChannel<ChannelType.PublicThread | ChannelType.PrivateThread | ChannelType.AnnouncementThread> {
```
vladfrangu, line 204: This feels wrong, thread channels are guild ones

deno/payloads/v10/channel.ts
```diff
@@ -145,7 +150,7 @@ export interface APIVoiceChannelBase<T extends ChannelType> extends APIGuildChan
 
 export interface APIGuildVoiceChannel
 	extends APIVoiceChannelBase<ChannelType.GuildVoice>,
-		Omit<APITextBasedChannel<ChannelType.GuildVoice>, 'name'> {
+		Omit<APITextBasedChannel<ChannelType.GuildVoice>, 'name' | 'last_pin_timestamp'> {
```
vladfrangu, line 153: Maybe this shouldn't extend TextBasedChannel instead?

vladfrangu (changes requested):

deno/payloads/v10/channel.ts
cyyynthia: I'm kind of unsure here. The fact messages can't be pinned in a text channel is unique to voice channels, so in my head it made sense to exclude it here specifically but keep it for all other channel types - avoiding duplicates on the way.

I can remove it from TextBasedChannel and add it individually to more specific types, if you prefer.

deno/payloads/v10/channel.ts
cyyynthia: Whoops you're right, I overlooked the type regression there from guild to basic 😅 - I'll make it extend both, as all the properties in `APIGuildTextChannel` are not applicable to threads.

vladfrangu (approved):

PR #748, opened 2023-04-14
Change snowflake typing

nk980113:
Make snowflake typing a little bit accurate.
Just make it a joke and close it if you want, but if you merge it, I'll appreciate.


nk980113:
Actually, there's a bug, but when I want to fix the bug, I found out that TypeScript throwed error to me

PR #756, opened 2023-04-22
feat: Migrate to eslint-config-neon

Jiralite:
In the website, I noticed the ESLint configuration was not a .json file. Is that intentional?

Also, the `lint:no-fix` rule might be wrong!

https://github.com/discordjs/discord-api-types/blob/3dac5b93e7568ba2fbd3bc30d229d2df80f96eed/website/package.json#L19-L20

Additionally, should it not check for .tsx files? Otherwise... this does not match anything as there appears to be no .ts or .mjs files.


.eslintrc.json
```diff
@@ -1,13 +1,20 @@
...
+	"extends": ["neon/common", "neon/node", "neon/typescript", "neon/prettier"],
 	"parser": "@typescript-eslint/parser",
 	"parserOptions": {
-		"project": "./tsconfig.eslint.json",
-		"extraFileExtensions": [".mjs"]
+		"project": "./tsconfig.eslint.json"
 	},
 	"plugins": ["local"],
 	"rules": {
+		"local/explicit-undefined-on-optional-properties": ["error", { "interfaceEndings": ["JSONBody"] }],
```
vladfrangu, line 9: This dupe seems wrong?

deno/payloads/v10/channel.ts
```diff
@@ -426,6 +426,7 @@ export enum ChannelType {
 	 *
 	 * @deprecated This is the old name for {@apilink ChannelType#AnnouncementThread}
 	 */
+	// eslint-disable-next-line @typescript-eslint/no-duplicate-enum-values
```
vladfrangu, line 429: Should we disable this rule globally?

vladfrangu (changes requested):
Also most likely, that script is wrong!

.eslintrc.json
Jiralite: Whilst I don't know the intricacies of the two local rules, both seem to exist in /.eslintplugin and seem to check for an explicit `undefined` on an optional property and a `?` on `undefined` properties respectively.

vladfrangu: Woops, misread the rule names 🥴

.eslintrc.json
```diff
@@ -1,13 +1,20 @@
...
 		"local/explicitly-optional-undefined-properties": ["error", { "interfaceEndings": ["JSONBody"] }],
-		"local/explicit-undefined-on-optional-properties": ["error", { "interfaceEndings": ["JSONBody"] }]
+		"@typescript-eslint/consistent-type-definitions": ["error", "interface"],
+		"@typescript-eslint/prefer-literal-enum-member": "off",
+		"@typescript-eslint/sort-type-union-intersection-members": "off",
+		"import/extensions": "off",
+		"tsdoc/syntax": "off",
```
vladfrangu, line 15: ho boy I sure hope this has autofix for when we eventually split the package up into multiple packages

deno/payloads/v10/channel.ts
Jiralite: It's only used in 2 places (excluding /deno). I think it may be fine for now.

vladfrangu (approved):

PR #762, opened 2023-04-29
fix: allow sending empty choices with autocomplete: true

almeidx:
https://github.com/discordjs/discord.js/blob/6229597db2be67be9e3aa0bf05574d826315eb36/packages/builders/__tests__/interactions/SlashCommands/Options.test.ts#L103-L106


deno/payloads/v10/_interactions/_applicationCommands/_chatInput/base.ts
```diff
@@ -22,6 +22,7 @@ export type APIApplicationCommandOptionWithAutocompleteOrChoicesWrapper<
 > =
 	| (Base & {
 			autocomplete: true;
+			choices?: never[];
```
vladfrangu, line 25: I'd instead mark this as just `[]`, does that not work?

PR #772, opened 2023-06-05
Add custom_id type for button component

ArDevy:
Added custom_id as a type to the base button typing, when using buttonBuilder.data.custom_id it would result in error. I do not know if there is something underlying this but this seems like the most applicable fix.

Cannot reference.


vladfrangu:
I will not be merging this. Buttons with the style of Link will never have custom_ids.
You also shouldn't just go into builder.data unless you know what you're doing! If you need to get a value from a builder that's known to create a button with a custom id, cast builder.data to APIMessageComponentButton (or similar)

ArDevy:
ahhhh, yes. I forgot about the link buttons. I will be looking into a change that would seem beneficial. Thanks.

PR #797, opened 2023-07-27
fix(Guild): union with never type

MARCROCK22:

The usage of 'never' in discord-api-types for all API responses with a 204 (empty) response leads to potential confusion among developers due to implicit precedence given to the declared type in TypeScript. It is essential to diligently avoid such undesirable situations.


sdanialraza (approved):

vladfrangu (approved):
Temporary stopgap for the issue, ty

PR #831, opened 2023-09-23
chore: remove deprecated importsNotUsedAsValues from tsconfig

### LINKED ISSUE #826: build(tsconfig): importsNotUsedAsValues deprecation
### Feature

discord-api-types has used tsconfig's `importsNotUsedAsValues: "error"` since 0.20.0. It stops ESModule-style imports being used without the type specifier; for instance:

```ts
import type { Snowflake } from '../globals';
```
is fine, but
```ts
import { Snowflake } from '../globals';
// This import is never used as a value and must use 'import type' because 'importsNotUsedAsValues' is set to 'error'. ts(1371)
```
gets rejected by the compiler.

`importsNotUsedAsValues` was [deprecated](https://github.com/microsoft/TypeScript/issues/51479) in TS 5.0, and will be removed in 5.5. Its replacement, [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax), provides different functionality: although it does cover the above import case, it also prevents top-level value exports in CommonJS modules:

```ts
import { Snowflake } from '../globals';
// Snowflake' is a type and must be imported using a type-only import when 'verbatimModuleSyntax' is enabled. ts(1484)

export const GatewayVersion = '10';
// A top-level 'export' modifier cannot be used on value declarations in a CommonJS module when 'verbatimModuleSyntax' is enabled. ts(1287)
```

### Ideal solution or implementation

Since moving to eslint-config-neon, type import behaviour is enforced by `@typescript-eslint/consistent-type-imports` anyway, so removing the flag entirely might be the easiest solution.

### Alternative solutions or implementations

_No response_

### Other context

_No response_

Renegade334:
Fixes #826

### DIFF
--- a/.eslintplugin/tsconfig.json
+++ b/.eslintplugin/tsconfig.json
@@ -1,19 +1,8 @@
 {
-	"compileOnSave": true,
+	"extends": "../tsconfig.json",
 	"compilerOptions": {
-		"alwaysStrict": true,
-		"lib": ["esnext"],
-		"module": "commonjs",
-		"noUnusedParameters": true,
-		"sourceMap": true,
 		"declaration": false,
-		"noUnusedLocals": true,
-		"removeComments": false,
-		"target": "ES2020",
-		"importsNotUsedAsValues": "error",
-		"strictNullChecks": true,
-		"preserveConstEnums": true,
-		"exactOptionalPropertyTypes": true,
+		"declarationMap": false,
 		"skipLibCheck": true
 	},
 	"include": ["./index.ts"]
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -11,7 +11,6 @@
 		"noUnusedLocals": true,
 		"removeComments": false,
 		"target": "ES2020",
-		"importsNotUsedAsValues": "error",
 		"strictNullChecks": true,
 		"preserveConstEnums": true,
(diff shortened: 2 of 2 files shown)


almeidx (approved):

vladfrangu (dismissed):

almeidx (changes requested):
On second thought, can you remove this one as well?
<https://github.com/discordjs/discord-api-types/blob/06fb47c51dc74cd9237f8328ac397590ad305d18/.eslintplugin/tsconfig.json#L13>

Renegade334:
> On second thought, can you remove this one as well?

would it be better to `extends`-ify it instead, like [tsconfig.eslint.json](/discordjs/discord-api-types/blob/main/tsconfig.eslint.json)?

vladfrangu:
I don't think that's necessary (I also don't think you can do that outside the eslint-plugins tsconfig file also ending up including all of the other files)

Renegade334:
@vladfrangu: the plugin's tsconfig will need updating anyway if it's using a deprecated feature. de634e8eb97ba17ba4d1eb528870a882c2fa1003 does compile the plugin with no issues.

almeidx (approved):

PR #849, opened 2023-11-17
fix(TextChannelType): Remove forum and media channels

Jiralite:
These are not text channel-like. It is the threads within these channels that are.

Removing these from the union brought about an error on `APIGuildTextChannel` as `APIThreadOnlyChannel` extended this. It contains the same properties needed for thread-based channels _and_ text channels, so I just added the forum and media channel types there. Maybe in the future, the class could be renamed.

### DIFF
--- a/deno/payloads/v10/channel.ts
+++ b/deno/payloads/v10/channel.ts
@@ -48,10 +48,8 @@ export type TextChannelType =
 	| ChannelType.PrivateThread
 	| ChannelType.AnnouncementThread
 	| ChannelType.GuildText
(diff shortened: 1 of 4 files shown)


suneettipirneni (approved):

vladfrangu (approved):

almeidx (approved):
We *could* make an interface with the common fields and extend it on both `APIGuildTextChannel` and `APIThreadOnlyChannel` to reduce duplication, but it's fine either way

suneettipirneni (approved):

PR #861, opened 2023-12-26
fix(GatewayThreadDispatch): properly type thread create/update/delete dispatches

### LINKED ISSUE #860: Broken types for thread events
### Issue description

The types for `GatewayThreadCreateDispatch`, `GatewayThreadUpdateDispatch`, and  `GatewayThreadDeleteDispatch` are wrong. They all point to `GatewayChannelModifyDispatch` instead of their own types, which causes behavior that doesn't match with Discord's documentation.

Interestingly, this appears to have been considered but not properly implemented, as evident by the existence of `GatewayThreadCreateDispatchData`. However, this type doesn't appear to be referenced anywhere.

### Code sample

```typescript
function handleEvent(event: GatewayDispatchPayload) {
  // narrow the event down
  if (event.t !== GatewayDispatchEvents.ThreadCreate) return;

  // expected behavior: this should work, newly_created is defined in the docs:
  // https://discord.com/developers/docs/topics/gateway-events#thread-create

  console.log(event.d.newly_created);
  //                  ^^^^^^^^^^^^^ 
  //                  Property 'newly_created' does not exist on type 'APIChannel'.



  // expected behavior: this should work, this event is supposed to give us a channel of type APIThreadChannel

  console.log(event.d.message_count);
  //                  ^^^^^^^^^^^^^ 
  //                  Property 'message_count' does not exist on type 'APIChannel'.
}
```

### Package version

discord-api-types@0.37.61

### Runtime

Node.js

### Runtime version

Node.js v21.4.0, TypeScript 4.9.5

High (immediate attention needed)

Saghetti0:
Closes #860

### DIFF
--- a/deno/gateway/v10.ts
+++ b/deno/gateway/v10.ts
@@ -34,6 +34,7 @@ import type {
 	AutoModerationRuleTriggerType,
 	APIAuditLogEntry,
 	APIEntitlement,
+	ChannelType,
 } from '../payloads/v10/mod.ts';
 import type { Nullable } from '../utils/internals.ts';
 
@@ -339,7 +340,9 @@ export type GatewayDispatchPayload =
 	| GatewayThreadListSyncDispatch
 	| GatewayThreadMembersUpdateDispatch
 	| GatewayThreadMemberUpdateDispatch
-	| GatewayThreadModifyDispatch
+	| GatewayThreadCreateDispatch
+	| GatewayThreadUpdateDispatch
+	| GatewayThreadDeleteDispatch
 	| GatewayTypingStartDispatch
 	| GatewayUserUpdateDispatch
 	| GatewayVoiceServerUpdateDispatch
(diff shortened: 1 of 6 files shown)


vladfrangu:
Please run `npm run clean` and then re-build deno types

Saghetti0:
Done, code now builds successfully and passes tests.

Jiralite (approved):

deno/gateway/v10.ts
```diff
@@ -1635,22 +1631,47 @@ export interface GatewayThreadCreateDispatchData extends APIThreadChannel {
...
+export type GatewayThreadDeleteDispatch = DataPayload<
+	GatewayDispatchEvents.ThreadDelete,
+	GatewayThreadDeleteDispatchData
+>;
 
 /**
  * https://discord.com/developers/docs/topics/gateway-events#thread-delete
  */
-export type GatewayThreadDeleteDispatchData = GatewayChannelModifyDispatchData;
+export interface GatewayThreadDeleteDispatchData {
```
vladfrangu, line 1655: This seems to be an extension of APIPartialChannel, let's reuse that please

vladfrangu (changes requested):

deno/gateway/v10.ts
Saghetti0: `APIPartialChannel` possibly has a name, whereas this event never does. Would it still be worth it to reuse that type, even though it's only for 2 fields (`id` and `type`)?

vladfrangu: oh. 😢 ok nvm me

vladfrangu:
Can you re-export GatewayThreadModifyDispatch as a deprecated type? Otherwise this is a breaking change

Saghetti0:
Alright, I added it back and put in a deprecation notice

Jiralite (approved):

almeidx (approved):

vladfrangu (approved):

PR #868, opened 2024-01-31
types(CDNRoutes): use generics for `format` options

### LINKED ISSUE #866: Route extension typings
### Issue description

`CDNRoutes` on API v10 has functions that return routes where most of them take a file format as a parameter (png, jpeg, etc.). The downside that this has caused is that the typings for them result in their return types being large template literal union types - one for each format possible. This not only makes it harder for those using these functions to quickly comprehend what will actually be returned from a given function, however it also from what I can see makes more work for those maintaining the typings for CDNRoutes - having to make sure the comment, allowed formats, and union return types are all in sync. I think this can be done better.

My proposal is to utilize generics for each of these CDNRoutes functions, which drastically reduces the amount of actual content needed to complete and maintain the typings for each function, and also makes for a smarter template literal return type which mutates based on the format specified - thus making for a better experience for those using these functions from my point of view.

I'm more than welcome to hearing others' thoughts on this and opening a PR assuming there's no issues with this that others are able to point out.

As a side note, the CDNRoutes `storePageAsset` function is implemented incorrectly. Its own [comments](https://github.com/discordjs/discord-api-types/blob/main/rest/v10/index.ts#L1113 ) as well as Discord's [documentation](https://discord.com/developers/docs/reference#image-formatting-cdn-endpoints) specify that it supports the PNG, JPEG, and WebP extensions, yet it's written to only return a PNG extension with no option to specify which available one to use.

### Code sample

```typescript
// Current behavior
import { CDNRoutes, ImageFormat } from "discord-api-types/v10";

CDNRoutes.emoji("id", ImageFormat.PNG);
// Return value typed as `/emojis/${string}.jpeg` | `/emojis/${string}.png` | `/emojis/${string}.webp` | `/emojis/${string}.gif`



// Ideal behavior
import { CDNRoutes, ImageFormat } from "discord-api-types/v10";

CDNRoutes.emoji("id", ImageFormat.PNG);
// Return value typed as `/emojis/${string}.png`
```

### Package version

0.37.61

### Runtime

Node.js

### Runtime version

v18.17.0 (TypeScript v5.2.2)

Low (slightly annoying)

TAEMBO:
Implements generics for `format` options. This removes the cumbersome template literal union typings that are returned, and allows those using CDNRoutes to more easily understand what's returned from a given route function. Retains strict boundaries for what extensions are allowed to be used for a given route.

close #866

### DIFF
--- a/deno/rest/v10/mod.ts
+++ b/deno/rest/v10/mod.ts
@@ -943,7 +943,7 @@ export const CDNRoutes = {
 	 *
 	 * This route supports the extensions: PNG, JPEG, WebP, GIF
 	 */
-	emoji(emojiId: Snowflake, format: EmojiFormat) {
+	emoji<TFormat extends EmojiFormat>(emojiId: Snowflake, format: TFormat) {
 		return `/emojis/${emojiId}.${format}` as const;
 	},
 
@@ -955,7 +955,7 @@ export const CDNRoutes = {
 	 *
 	 * This route supports the extensions: PNG, JPEG, WebP, GIF
 	 */
-	guildIcon(guildId: Snowflake, guildIcon: string, format: GuildIconFormat) {
+	guildIcon<TFormat extends GuildIconFormat>(guildId: Snowflake, guildIcon: string, format: TFormat) {
 		return `icons/${guildId}/${guildIcon}.${format}` as const;
 	},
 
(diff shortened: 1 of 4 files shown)


vladfrangu (dismissed):

sdanialraza:
Discussed this internally, you don't need to prefix the type parameters with T, so just `Format` or `Index` should be good

deno/rest/v10/mod.ts
```diff
@@ -1130,7 +1151,7 @@ export const CDNRoutes = {
 	 *
```
vladfrangu, line 1151: Need to also apply this to storePageAsset

vladfrangu (changes requested):

vladfrangu (dismissed):

almeidx (dismissed):

vladfrangu (approved):

almeidx (approved):

PR #998, opened 2024-06-04
fix(APIGuildMember): make user required and omit in messages

advaith1:
Currently `user` is optional on APIGuildMember, but the only case it isn't present is in the message gateway events, because those have message.author so message.member.user would be redundant. This leads to a bad experience when using other member objects, since `user` will be there but the optionality needs to be handled.

This PR sets `user` required in APIGuildMember, and omits it where it is not present (GatewayMessageEventExtraFields).


vladfrangu (changes requested):
Please name the test file in the same pattern as the other files (so name.test-d.ts)

advaith1:
oops i thought i did that. fixed

almeidx (approved):

PR #1046, opened 2024-08-07
feat(Routes): voice state endpoint

imnaiyar:
 
 - https://github.com/discord/discord-api-docs/pull/7061


imnaiyar:
Also I'm not sure if we should rename `GatewayVoiceState` type (possibly to `APIVoiceState` ?) now that it is possible to get it over http

imnaiyar:
> > Also I'm not sure if we should rename `GatewayVoiceState` type (possibly to `APIVoiceState` ?) now that it is possible to get it over http
> 
> Yes, and deprecating the gateway alias

Did that if you wanna have a look

rest/v10/index.ts
```diff
@@ -609,6 +609,15 @@ export const Routes = {
 		return `/voice/regions` as const;
 	},
 
+	/**
+	 * Route for:
+	 * - GET `/guilds/{guild.id}/voice-states/@me`
+	 * - GET `/guilds/{guild.id}/voice-states/{user.id}`
+	 */
+	voiceState(guildId: Snowflake, userId: Snowflake | '@me' = '@me') {
```
TAEMBO, line 617: Shouldn't we be reusing the existing [`guildVoiceState()`](https://github.com/discordjs/discord-api-types/blob/main/rest/v10/index.ts#L793) function for this instead? That function uses the same route as this, and would only require additional JSDoc comments to indicate it now supports the GET method as well.

imnaiyar: You're right, I missed this. Though voice state endpoints have been moved to a separate category altogether (moved from Guild) in the API documentation, so I'm wondering if we should still call it `guildVoiceState`

imnaiyar: Ig it doesn't make much sense renaming it.

TAEMBO (approved):

almeidx (changes requested):
Missing the REST types, which should be added to rest/voice.ts to match the Discord docs

The patch endpoint types are in rest/guild.ts, which should be moved to rest/voice.ts as well. Not sure if we'd have to add a re-export on the patch types in rest/guild.ts since users *can* import individual files too

imnaiyar:
> Missing the REST types, which should be added to rest/voice.ts to match the Discord docs
> 
> The patch endpoint types are in rest/guild.ts, which should be moved to rest/voice.ts as well. Not sure if we'd have to add a re-export on the patch types in rest/guild.ts since users _can_ import individual files too

I'm not sure if we should re-export from

> Missing the REST types, which should be added to rest/voice.ts to match the Discord docs
> 
> The patch endpoint types are in rest/guild.ts, which should be moved to rest/voice.ts as well. Not sure if we'd have to add a re-export on the patch types in rest/guild.ts since users _can_ import individual files too

not re-exporting from guild.ts right now, we can wait for others to see what they want to do

vladfrangu:
> since users _can_ import individual files too

No, they cannot. So moving exports from the right file is perfectly fine

vladfrangu (changes requested):

vladfrangu (approved):

PR #1065, opened 2024-08-19
refactor(rest): ensure types follow naming pattern

almeidx:
Added a custom ESLint rule to ensure REST types follow the following naming pattern:

```js
/^REST(?:Get|Patch|Post|Put|Delete)\w+(?:JSONBody|FormDataBody|URLEncodedData|Result|Query)$/
```


.eslint-plugin-local/index.ts
```diff
@@ -131,5 +131,57 @@ export = {
...
+								messageId: 'invalidName',
+								data: { name },
+							});
+						}
+					},
+				};
+			},
+			meta: {
+				messages: {
+					invalidName: '{{ name }} does not match REST type naming convention',
```
vladfrangu, line 167: Lets add in the naming convention somehow so users know ahead of time

.eslint-plugin-local/index.ts
```diff
@@ -37,6 +37,9 @@ const schema = [
 	},
 ] as const;
 
+const REST_TYPE_NAME_REGEX =
+	/^REST(?:Get|Patch|Post|Put|Delete)\w+(?:JSONBody|FormDataBody|URLEncodedData|Result|Query)$/;
```
Syjalo, line 41: `\w` includes `_`, which is not allowed in PascalCase
```suggestion
	/^REST(?:Get|Patch|Post|Put|Delete)[a-zA-Z0-9]+(?:JSONBody|FormDataBody|URLEncodedData|Result|Query)$/;
```

Syjalo (changes requested):

vladfrangu (approved):

PR #1101, opened 2024-09-17
fix(rest/oauth2): correct string literal types containing bot scope

Renegade334:

Minor fixes for the `scope` property type unions within the bot-specific query interfaces in rest/oauth2:
- Wrap the bare case (`OAuth2Scopes.Bot`) in a template literal, so that passing `scope: "bot"` is valid and doesn't raise a compilation error.
- No longer consider `%20` a valid space. For the parameters to exist in object form, they are by definition not yet URL-encoded, and passing a property like `scope: "bot%20identify"` to URLSearchParams would just re-encode the percent sign, resulting in an invalid OAuth2 query.
- Fix the order of the "sandwich" case (`"activities.read bot identify"`). Previously, this string would raise a compilation error, due to the misplaced final space in the template literal type.

### DIFF
--- a/deno/rest/v10/oauth2.ts
+++ b/deno/rest/v10/oauth2.ts
@@ -134,10 +134,10 @@ export interface RESTOAuth2BotAuthorizationQuery {
 	 * Needs to include bot for the bot flow
 	 */
 	scope:
-		| OAuth2Scopes.Bot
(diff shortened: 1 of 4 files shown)


vladfrangu (approved):
Can you add in a tsd test too? Similar to the one did for routes

almeidx (approved):

PR #1155, opened 2024-11-22
refactor(GatewayThreadListSync): Narrow `threads`

Jiralite:
Narrowed the `threads` property to only contain threads.


vladfrangu (approved):
How was this not spotted sooner

PR #645, opened 2022-11-17
feat(rpc): add *RPCCommands and RPCEvents* 

TrueXPixells:
**Please describe the changes this PR makes and why it should be merged:**
i added more rpc commands and events
Note: after /* */ comments are not including discord rpc docs

PR #529, opened 2022-07-15
feat: gamesdk support

vvito7:
**Please describe the changes this PR makes and why it should be merged:**
This PR adds support for all GameSDK stuff.
I ran into many problems while writing payloads and rest files, such as doubts about [Update User Achievement](https://discord.com/developers/docs/game-sdk/achievements#update-user-achievement) (this one seems not to be working) and [Get User Achievements](https://discord.com/developers/docs/game-sdk/achievements#get-user-achievements) (this one I just don't understand) endpoints. So be free to correct or explain me some mistake I made 😄 
Closes #68 and replaces #88.

**If applicable, please reference Discord API Docs PRs or commits that influenced this PR:**

PR #1200, opened 2025-01-26
feat(RPC): types

OfficialSirH:

Closes #645 
Closes #529 


rpc/common.ts
```diff
@@ -1,33 +1,508 @@
+import type { Snowflake } from '../globals';
+import type { APIMessage, APIUser } from '../v10';
+
+/**
+ * @unstable the ping object for the `VOICE_CONNECTION_STATUS` dispatched `pings` field,
+ * but discord's documentation incorrectly documents it as an 'array of integers'.
```
Jiralite, line 6: This is probably something to raise to Discord.

rpc/common.ts
```diff
@@ -1,33 +1,508 @@
...
+ */
+export interface RPCVoiceConnectionStatusPing {
+	/**
+	 * the time the ping was sent
+	 */
+	time: number;
+	/**
+	 * the latency of the ping (in ms)
```
Jiralite, line 14:
```suggestion
	 * The latency of the ping in milliseconds
```

rpc/common.ts
```diff
@@ -1,33 +1,508 @@
...
+/**
+ * @unstable
+ */
+export interface RPCAPIMessageParsedContentMention {
+	type: 'mention';
+	userId: Snowflake;
+	channelId: Snowflake;
+	guildId: Snowflake;
+	/**
+	 * same as `userId`
```
Jiralite, line 45:
```suggestion
	 * Same as {@link RPCAPIMessageParsedContentMention.userId}
```

rpc/common.ts
```diff
@@ -1,33 +1,508 @@
...
+	 */
+	parsedUserId: Snowflake;
```
Jiralite, line 47: Might be better to type this as `RPCAPIMessageParsedContentMention["userId"]` if it truly is the same.

rpc/v10.ts
```diff
@@ -1 +1,2606 @@
...
+}
+
+/**
+ * https://discord.com/developers/docs/topics/rpc#setvoicesettings-set-voice-settings-argument-and-response-structure
+ */
+export type RPCSetVoiceSettingsResultData = RPCGetVoiceSettingsResultData;
+/**
+ * https://discord.com/developers/docs/topics/rpc#setvoicesettings-set-voice-settings-argument-and-response-structure
+ *
+ * @note Discord only supports a single modifier of voice settings at a time over RPC. If an app changes voice settings, it will lock voice settings so that other apps connected simultaneously lose the ability to change voice settings. Settings reset to what they were before being changed after the controlling app disconnects. When an app that has previously set voice settings connects, the client will swap to that app's configured voice settings and lock voice settings again.
```
Jiralite, line 597:
```suggestion
 * @remarks Discord only supports a single modifier of voice settings at a time over RPC. If an app changes voice settings, it will lock voice settings so that other apps connected simultaneously lose the ability to change voice settings. Settings reset to what they were before being changed after the controlling app disconnects. When an app that has previously set voice settings connects, the client will swap to that app's configured voice settings and lock voice settings again.
```

Jiralite (changes requested):
1. Can you go through all the documentation and capitalise the first letter?
2. "rpc" should be capitalised in the documentation everywhere. Would you mind doing that?
3. When you add hyperlinks, do so like `{@link <link> | <display_text>}` instead of `[<text>](<link>)`

rpc/common.ts
Jiralite: A lot of errors? Oh dear...

rpc/v10.ts
```diff
@@ -1 +1,2606 @@
...
+
+/**
+ * https://discord.com/developers/docs/topics/rpc#setuservoicesettings-set-user-voice-settings-argument-and-response-structure
+ */
+export type RPCSetUserVoiceSettingsResultData = Required<RPCSetUserVoiceSettingsArgs>;
+
+/**
+ * https://discord.com/developers/docs/topics/rpc#selectvoicechannel-select-voice-channel-argument-structure
+ *
+ * @warning When trying to join the user to a voice channel, you will receive a `5003` error coded response if the user is already in a voice channel. The `force` parameter should only be specified in response to the case where a user is already in a voice channel and they have approved to be moved by your app to a new voice channel.
```
sdanialraza, line 456:
```suggestion
 * @warning When trying to join the user to a voice channel, you will receive
 * a {@link RPCErrorCodes.SelectVoiceForceRequired} error coded response if the user is already in a voice channel.
 * The `force` parameter should only be specified in response to the case where a user is already in a voice channel
 * and they have approved to be moved by your app to a new voice channel.
```

rpc/v10.ts
```diff
@@ -1 +1,2606 @@
...
+ */
+export interface RPCSetActivityArgs {
+	/**
+	 * The application's process id
+	 */
+	pid: number;
+	/**
+	 * The rich presence to assign to the user
+	 */
+	activity?: Partial<
```
sdanialraza, line 585:
```suggestion
	activity?: Partial<Omit<GatewayActivity, 'created_at' | 'id'>>
```
Also, the `type` can be narrowed further but not sure if we want to do that
image

rpc/v10.ts
```diff
@@ -1 +1,2606 @@
...
+	/**
+	 * Guild with requested id
+	 */
+	guild: APIPartialGuild;
+	/**
+	 * Number of online users in guild
+	 *
+	 * @deprecated This will always be 0
+	 */
+	online: number;
```
sdanialraza, line 1442:
```suggestion
	online: 0;
```

sdanialraza (changes requested):

rpc/v10.ts
Jiralite: Do not think we use `@warning`. I believe this should be a `@remarks`? Maybe @almeidx would know better.

almeidx: Yes, it should use `@remarks`

sdanialraza (approved):

PR #1282, opened 2025-06-26
chore: fix weird issue with api-extractor

Qjuh:

For whatever reasons (tell me if you know) the removal of the explicit `/index´ caused `api-extractor` to error with

```js
Error: Internal Error: Cannot assign isExternal=true for the symbol APIEmoji because it was previously registered with isExternal=false
```


almeidx (approved):
I'm in favour of keeping the `/index` suffix anyway (only missing file extensions next 😄)

vladfrangu (approved):
Lol, you fixed deno build with this. Ty

PR #1307, opened 2025-07-22
Added Route guildMembersSearchV2 for POST /guilds/{guild.id}/members-search

Lixqa:
This PR adds support for the POST /guilds/{guild.id}/members-search route, which enables searching for guild members using a variety of query parameters - not just by nickname as with the older endpoint.

This is a preparatory step toward reworking GuildMemberManager#search() in discord.js to utilize this newer, more flexible route. By doing so, we can offer developers greater control and precision when searching for guild members.

The name of the route includes "V2", even if it's not ideal. This just follows the route naming 🤷‍♂️.

/

### DIFF
--- a/rest/v10/index.ts
+++ b/rest/v10/index.ts
@@ -313,6 +313,14 @@ export const Routes = {
 
 	/**
 	 * Route for:
+	 * - POST `/guilds/{guild.id}/members-search`
+	 */
+	guildMembersSearchV2(guildId: Snowflake) {
+		return `/guilds/${guildId}/members-search` as const;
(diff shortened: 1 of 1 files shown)


vladfrangu:
This endpoint isn't documented on Discord's end right? If not, I'm torn on documenting it, and if we do, it MUST be documented with `@unstable`.

Lixqa:
No it isn't documented. It didn't even worked for bots like 3 months ago when I tested it last time. Seems like they enabled it for bots now. Maybe they are working on documenting it right now.

PR #1526, opened 2026-02-16
fix(Types): replace never usage with undefined

### LINKED ISSUE #1522: HTTP 204 route result types should be void instead of never
### Issue description

The implications of the never type means you will never be able to observe the type which isn't true in the case of 204 as those are usually represented through undefined. 204 is an OK status code, so throwing an Exception to satisfy the semantics of never doesn't make sense.

In the [Typescript handbook](https://www.typescriptlang.org/docs/handbook/2/functions.html#never), they describe it as appropriate for use in methods which always return an exception as you cannot observe the value even in try catch blocks. The only exception to this case I can think of is process.exit() being typed as returning never, but that also makes sense as you cannot observe that return type.

In practice, this requires extra code to cast as a route result type even if the return of making the request through some other function is any. Users of my lib that implements these types have also complained that it makes writing mocks for the real functions annoying, or even impractical in the case of JS without writing a bunch of type annotations or ts-ignore directives over each offending entry. Attached is an image of such a case

Image

### Code sample

```typescript
// From the library as an example of needing to cast
// The return type of RequestHandler.request is Promise<any> which the semantics of any
// implies it can be ANYTHING, but it errors with never and nothing else.
public async leaveGuild(guildId: string): Promise<RESTDeleteAPICurrentUserGuildResult> {
	return this.requestHandler.request(Endpoints.USER_GUILD("@me", guildId), {}, "delete") as RESTDeleteAPICurrentUserGuildResult;
}
```

### Package version

0.38.38

### Runtime

Node.js

### Runtime version

v24.13.0

Low (slightly annoying)

PapiOphidian:
closes #1522

The majority of the usage of `never` in the project does not align with the semantics of the keyword and this PR replaces them with `void`. The keyword implies the value being typed as `never` cannot be observed under any conditions such as in the case of only throwing an exception or `process.exit()`, which is not true in cases of where these types would be used such as return types for routes that return an HTTP 204 which is an OK status code.
It's also used in some gateway payloads which is also technically incorrect as attempting to access the values at runtime would produce an observable `undefined`.

Some of the base types for gateway payloads that would have been modified would not work to be replaced as void, so those have been kept.
API v6-v8 also has not been modified as those had clear deprecation comments, but it was assumed that api v9 was still being maintained, so types from v9 have also been modified.

Some markdown mentioning the usage of never has also been modified.

N.A.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ The exports of each API version is split into three main parts:
         - Some exported types (specifically OAuth2 related ones) may not respect this entire structure due to the nature of the fields. They will start with either `RESTOAuth2` or with something similar to `REST<HTTP Method>OAuth2`
 
     - If a type ends with `Result`, then it represents the expected result by calling its accompanying route.
-        - Types for a Result that will be a `204 No Content` will be typed as `void`
+        - Types for a Result that will be a `204 No Content` will be typed as `undefined`
 
 - Anything else that is miscellaneous will be exported based on what it represents (for example the `REST` route object).
 
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ The exports of each API version is split into three main parts:
         - Some exported types (specifically OAuth2 related ones) may not respect this entire structure due to the nature of the fields. They will start with either `RESTOAuth2` or with something similar to `REST<HTTP Method>OAuth2`
 
     - If a type ends with `Result`, then it represents the expected result by calling its accompanying route.
-        - Types that are exported as `never` usually mean the result will be a `204 No Content`, so you can safely ignore it. This does **not** account for errors.
+        - Types for a Result that will be a `204 No Content` will be typed as `void`
 
 - Anything else that is miscellaneous will be exported based on what it represents (for example the `REST` route object).
 
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ The exports of each API version is split into three main parts:
         - Some exported types (specifically OAuth2 related ones) may not respect this entire structure due to the nature of the fields. They will start with either `RESTOAuth2` or with something similar to `REST<HTTP Method>OAuth2`
 
     - If a type ends with `Result`, then it represents the expected result by calling its accompanying route.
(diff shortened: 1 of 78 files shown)


vladfrangu:
> Some of the base types for gateway payloads that would have been modified would not work to be replaced as void, so those have been kept

Which ones? 👁️

vladfrangu:
Also should we use `undefined` instead of void?

Jiralite:
I think for types like this one:

`export type RPCSubscribeEntitlementDeleteArgs = Record<string, void>;`

`undefined` might fit better. It's like saying an object comprises solely of keys?

vladfrangu:
I'm confused what that type even is, cuz surely its not an object of keys...

PapiOphidian:
It was the heartbeat and reconnect packets that I couldn't change to void without some ts errors. I hadn't considered undefined, but I do believe undefined could work better and it addresses the issue of those packets not being able to be changed from never. I've updated the usages I'd previously updated as well as documentation

vladfrangu:
Can you please rebase your PR and adjust to undefined instead of void?

PapiOphidian:
I've rebased the branch to upstream main and I've already changed it to use undefined instead of void prior

almeidx:
Seems like there were some issues with your rebase

PapiOphidian:
Sorry about that. First time needing to rebase

almeidx:
This endpoint needs to be changed too:
https://github.com/PapiOphidian/discord-api-types/blob/b5ae2d9da718bcd7bc5a2b25847bb7983708039a/rest/v10/soundboard.ts#L7

It was updated in https://github.com/discordjs/discord-api-types/pull/1568

vladfrangu (dismissed):

almeidx (dismissed):
- `RESTPostAPISendSoundboardSoundResult` change is missing on v9
- `client_id?: never` change is missing on v10
- errors callout missing on README

almeidx (approved):

Jiralite (approved):

PR #1625, opened 2026-04-14
fix(APITeamMember): `permissions` is now missing

Jiralite:
At or before 2026-04-14, Discord stopped sending this field.

I have not opted to remove the field, but would be open to it since this is supposed to mirror the Discord API.


Jiralite:
`permissions` is coming back.[^1]

[^1]: https://canary.discord.com/channels/222078108977594368/999796393810067607/1493671429777129503

PR #1705, opened 2026-06-28
feat: support `None` enum value for flags

### LINKED ISSUE #1699: Convert all enumerated types to plain objects

Qjuh:

Passing `flags: 0` was not allowed with current bitfield Flags enums, adding a `None = 0` value to each of them solves this issue. Downstream users need to make sure to not handle these as actual flags but just as placeholder for no flags.

Closes #1699

### DIFF
--- a/payloads/v10/gateway.ts
+++ b/payloads/v10/gateway.ts
@@ -406,6 +406,7 @@ export interface GatewayActivitySecrets {
  * @see {@link https://discord.com/developers/docs/topics/gateway-events#activity-object-activity-flags}
  */
 export enum ActivityFlags {
+	None = 0,
 	Instance = 1 << 0,
 	Join = 1 << 1,
 	Spectate = 1 << 2,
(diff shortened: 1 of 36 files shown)


payloads/v10/gateway.ts
```diff
@@ -406,6 +406,7 @@ export interface GatewayActivitySecrets {
  * @see {@link https://discord.com/developers/docs/topics/gateway-events#activity-object-activity-flags}
  */
 export enum ActivityFlags {
+	None = 0,
```
vladfrangu, line 409: No comment here but others have it? I don't think we need a comment in general to explain what none means

Qjuh: I added comments to the enums that had comments for their other values, for consistency.

vladfrangu: let's just drop comments everywhere, it's pretty self explanatory what None means :D

PR #1769, opened 2026-08-08
fix: allow DM channels on ChannelOption channel_type

Qjuh:

ChannelOptions on ChatInputApplicationCommands actually allow `ChannelType.DM` and `ChannelType.GroupDM` in `channel_types` since they can be run in those kinds of channels as user  installed apps.

See https://github.com/discordjs/discord.js/pull/11380 for details


vladfrangu (changes requested):
v9

Qjuh:
> v9

This type doesn't even exist on v9.

PR #1784, opened 2026-08-30
fix(chatInput): disallow required on subcommand and subcommand group options

### LINKED ISSUE #1510: SubcommandOption types have a required property
### Issue description

try to use type `APIApplicationCommandSubcommandOption` or `APIApplicationCommandSubcommandGroupOption`

see that they have a `required` boolean property instead of being never
Discord Docs: https://discord.com/developers/docs/interactions/application-commands#:~:text=all%20but,SUB_COMMAND_GROUP

Image

### Code sample

```typescript
const testSubcommandOption: APIApplicationCommandSubcommandOption = {
	description: 'test',
	name: 'test',
	type: ApplicationCommandOptionType.Subcommand,
	options: [],
	required: true
}
const testSubcommandGroupOption: APIApplicationCommandSubcommandGroupOption = {
	name: 'testGroup',
	description: 'test group',
	type: ApplicationCommandOptionType.SubcommandGroup,
	options: [],
	required: false
}
```

### Package version

0.38.38

### Runtime

Node.js

### Runtime version

v23.10.0

Low (slightly annoying)

Willfric:
fixes #1510

`APIApplicationCommandSubcommandOption` and `APIApplicationCommandSubcommandGroupOption` both extend `APIApplicationCommandOptionBase` which includes `required?: boolean`. discord's docs explicitly state subcommands and subcommand groups do not support `required`, so both types incorrectly allow it.

override with `required?: never` on both interfaces to match the api spec.

### DIFF
--- a/deno/payloads/v10/_interactions/_applicationCommands/_chatInput/attachment.ts
+++ b/deno/payloads/v10/_interactions/_applicationCommands/_chatInput/attachment.ts
@@ -1,10 +1,10 @@
 import type { Snowflake } from '../../../../../globals.ts';
-import type { APIApplicationCommandOptionBase, APIInteractionDataOptionBase } from './base.ts';
+import type { APIApplicationCommandBasicOptionBase, APIInteractionDataOptionBase } from './base.ts';
 import type { ApplicationCommandOptionType } from './shared.ts';
 
 export type FileUploadType = 'audio' | 'image' | 'video' | `.${string}`;
 
-export interface APIApplicationCommandAttachmentOption extends APIApplicationCommandOptionBase<ApplicationCommandOptionType.Attachment> {
+export interface APIApplicationCommandAttachmentOption extends APIApplicationCommandBasicOptionBase<ApplicationCommandOptionType.Attachment> {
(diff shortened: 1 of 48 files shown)


almeidx (changes requested):
Please make these changes to v9 and also run the `build:deno` script

almeidx (dismissed):

Jiralite:
Why `never` instead of omitting them from the interface in the first place?

Willfric:
> Why `never` instead of omitting them from the interface in the first place?

can't omit it, it's inherited from `APIApplicationCommandOptionBase`. `never` is the only way to block an inherited property in ts

Jiralite:
So remove it there instead?

Willfric:
> So remove it there instead?

That'd break all the other types that extend the same base `string`, `integer`, `number`, etc since they all need `required`. `never` on just the subcommand interfaces

Jiralite:
And then again, my response would be the same in spirit: make a new shared base then?

Willfric:
So something like a base without `required` that subcommands extend, and a separate extended base with `required` for the rest?

